PR #11, opened 2024-10-21
Ddfbra 90 kort i sogeresultat grid

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-90

#### Description

This PR contains a result page with filters with initial setup for UI components.

Please refer to the tickets in the current sprint + TODOs in the code for the known deficiencies.


components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+      return acc
+    },
+    {} as { [key: string]: keyof SearchFilters[] }
+  )
+
+  const { data, isLoading } = useSearchWithPaginationQuery({
+    q: { all: q },
+    offset: 0,
+    limit: 10,
```
spaceo, line 82: Configuration comment as previous ones goes here...

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+          [mapFacetsToFilters[facetDefinition]]: [...values],
+        }
+      }
+
+      return acc
+    },
+    {} as { [key: string]: keyof SearchFilters[] }
+  )
+
+  const { data, isLoading } = useSearchWithPaginationQuery({
```
spaceo, line 79: I think we should not execute the query if we have no `q` value.
It should be possible to control that via the `enabled` property given to the hook.

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+    facets: facetDefinitions,
+    filters: {
+      branchId: branchIds,
+      ...facetsForSearchRequest,
+    },
+  })
+
+  return (
+    <div className="content-container">
+      <h1 className="mt-[88px] text-typo-heading-2">{`Viser resultater for "${q}" ${data?.search.hitcount ? "(" + data?.search.hitcount + ")" : ""}`}</h1>
```
spaceo, line 101: I think there is a bit too much string concatenation going on here. In my opinion we could make a small component for showing this line

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+      {isLoadingFacets && <p>isLoadingFacets...</p>}
+      {!dataFacets?.search?.facets?.length && <p>Ingen filter</p>}
+      {dataFacets?.search?.facets && dataFacets?.search?.facets?.length > 0 && (
```
spaceo, line 104: I am not fond of these long chains. Can't we create a variable that is called facets before rendering and the we can use that?

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <SearchFilterBar facets={dataFacets.search.facets} />
+      )}
+      {isLoading && <p>isLoading...</p>}
+      {data?.search.hitcount === 0 && <p>Ingen søgeresultat</p>}
+      {data?.search?.works && <SearchResults works={data.search.works} />}
```
spaceo, line 109: Same comment as above. And why is `data?.search.hitcount` not optional when `data?.search?.works` is?

components/pages/searchPageLayout/SearchResults.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import WorkCard from "@/components/shared/workCard/WorkCard"
+import { SearchWithPaginationQuery } from "@/lib/graphql/generated/fbi/graphql"
+
+type SearchResultProps = {
+  works: SearchWithPaginationQuery["search"]["works"]
+}
+
+const SearchResults = ({ works }: SearchResultProps) => {
+  // const router = useRouter()
```
spaceo, line 14: We dont need this, I guess?

components/shared/searchFilters/SearchFilterBar.tsx
```diff
@@ -0,0 +1,36 @@
+"use client"
+
+import React, { useState } from "react"
+
+import { SearchFacetsQuery } from "@/lib/graphql/generated/fbi/graphql"
+import SearchFiltersColumn from "./SearchFiltersColumn"
+
+type SearchFilterBarProps = {
+  facets: SearchFacetsQuery["search"]["facets"]
```
spaceo, line 9: Don't we have a fragment for the facets. Otherwise I think we should create one and use that directly.

components/shared/searchFilters/SearchFiltersColumn.tsx
```diff
@@ -0,0 +1,90 @@
+import { SearchFacetsQuery } from "@/lib/graphql/generated/fbi/graphql"
+import { cn } from "@/lib/utils"
+import React, { useEffect, useRef, useState } from "react"
+import { mapFacetsToFilters, mapFilterNameToTranslation, toggleFilter } from "./helper"
+import { useRouter, useSearchParams } from "next/navigation"
+import Icon from "../icon/Icon"
+import { Button } from "../button/Button"
+
+type SearchFiltersColumnProps = {
+  facet: SearchFacetsQuery["search"]["facets"][0]
```
spaceo, line 10: Don't we have a fragment for the facets. Otherwise I think we should create one and use that directly.

components/shared/searchFilters/SearchFiltersColumn.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const facetName = facet.name as keyof typeof mapFacetsToFilters
+  const searchParams = useSearchParams()
+  const elementRef = useRef<HTMLDivElement | null>(null)
+  const [hasOverflow, setHasOverflow] = useState(false)
+
+  useEffect(() => {
+    const el = elementRef.current
+    if (el) {
+      const isOverflowing = el.scrollHeight > el.clientHeight || el.scrollWidth > el.clientWidth
+      isOverflowing ? setHasOverflow(true) : null
```
spaceo, line 32: I think it us confusing to use ternaries when you use them for other purposes that assignments/value resolving.
I think this should be:
```typescript
if (isOverflowing) {
  setHasOverflow(true)
}

File: components/shared/searchFilters/helper.ts [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+} as Record<FacetField, keyof SearchFilters>
+
+export const mapFilterNameToTranslation: Partial<Record<keyof typeof mapFacetsToFilters, string>> =
+  {
+    materialTypesGeneral: "Type",
+    mainLanguages: "Sprog",
+    age: "Alder",
+    lix: "Lix",
+    subjects: "Emne",
+  }
```
spaceo, line 27: We should handle these facets/filters in a cetralized configuration manner. But let's talk about it and a `TODO`? would be in place...

components/shared/searchFilters/helper.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const toggleFilter = (filterName: string, value: string, router: AppRouterInstance) => {
+  const searchParams = new URLSearchParams(window.location.search)
```
spaceo, line 30: Although this is fine, I think it should be possible to get the query string from the `router`object.

components/shared/searchInput/SearchInput.tsx
```diff
@@ -1,35 +1,72 @@
...
+    }
+  }
+
+  const navigateToSearch = () => {
+    if (!inputRef.current) return
+    const inputValue = inputRef.current.value
+    router.push(inputValue ? `/search?q=${inputValue}` : "/search", {
+      scroll: false,
+    })
   }
```
spaceo, line 45: Is this accessible?
Have you tested the interface with a static A11y test or with a screen reader?
We need to be sharp on this 😅

components/shared/workCard/WorkCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+const WorkCard = ({ work }: WorkCardProps) => {
+  return (
+    <div className="mb-4">
+      <Link href={`/work/${work.workId}`}>
+        <div key={work.workId} className="rounded-sm bg-background-overlay p-2 md:p-4">
+          <div
+            className="px-auto shadow-coverPicture relative mx-auto mb-3 mt-6 flex aspect-[166/228] w-[calc(100%-76px-4px)]
+              items-center rounded-sm md:mb-6 md:mt-9 md:w-[calc(100%-106px)]">
+            <Image
+              src="https://i.pinimg.com/564x/b0/4a/b2/b04ab2de27591aee4a76015e42e0e282.jpg"
```
spaceo, line 23: If we use demo content like this we need to write a `TODO`about that real values should be used in the future.

components/shared/workCard/WorkCardAvailabilityRow.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react"
+
+import { SearchWithPaginationQuery } from "@/lib/graphql/generated/fbi/graphql"
+
+import { isOfWorkTypeCategory } from "./helper"
+import WorkCardAvailabilityItem from "./WorkCardAvailabilityItem"
+
+type WorkCardAvailabilityRowProps = {
+  materialTypes: SearchWithPaginationQuery["search"]["works"][0]["materialTypes"]
```
spaceo, line 9: Can we use a fragment?

components/shared/workCard/helper.ts
```diff
@@ -0,0 +1,46 @@
...
+    // We shorten to max
+    if (index === amount) {
+      return acc + ", et. al"
+    }
+    if (index > amount) {
+      return acc
+    }
+    return acc + (index > 0 ? ", " : "") + creator.display
+  }, "")
+}
```
spaceo, line 18: It is very important that this is consistent with how it has been done in dpl-react. Please double-check it. If you need a pointer to the place in the repository let me know.

spaceo (changes requested):
Great stuff. 🎉 
Please look at my comments.

app/search/fetchSearchResult.server.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+
+  return queryClient
+}
+
+const prefetchSearchFacets = async (q: string, queryClient: QueryClient) => {
+  await queryClient.prefetchQuery({
+    queryKey: useSearchFacetsQuery.getKey({
+      q: { all: q },
+      facetLimit: 100,
```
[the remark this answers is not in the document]
Adamik10: Yeah I agree that this shouldn't be configurable.

app/search/fetchSearchResult.server.ts
```diff
@@ -0,0 +1,43 @@
...
+      facets: ["materialTypesGeneral", "mainLanguages", "age", "lix", "subjects"] as FacetField[],
```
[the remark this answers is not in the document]
Adamik10: I agree :) it should be using a reference to a definition that is used all around the app

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -0,0 +1,114 @@
...
+  "751025",
+  "775133",
+  "751024",
+  "775100",
+  "775170",
+  "775150",
+  "775130",
+] as string[]
+
+export type FilterItemTerm = Omit<FacetValue, "__typename">
```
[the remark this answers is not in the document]
Adamik10: I am assuming it might have been just cosmetic 🤷

components/pages/searchPageLayout/SearchResults.tsx
Adamik10: You're right :) We don't

components/shared/searchFilters/SearchFilterBar.tsx
Adamik10: Great thought!

components/shared/searchFilters/SearchFiltersColumn.tsx
Adamik10: Fair enough. I was coming from a place where I didn't want to nest if statements

components/shared/searchFilters/helper.ts
Adamik10: I'm not sure if these are a part of configuration. I don't think they ever should be configurable.

components/shared/searchInput/SearchInput.tsx
Adamik10: Great point!!! :)) but yes, it is indeed accessible by keyboard. I also added aria labels around the app now that you mentioned accessibility :)

components/shared/workCard/helper.ts
Adamik10: It is the same, but dynamic in how many we want to shorten it to. In this case I chose two, so that the search results look good + children probably wouldn't be able to soak in more information than two names.

components/shared/searchFilters/helper.ts
spaceo: Did you follow up on this? You did not reply :)

components/shared/workCard/helper.ts
spaceo: Check. Would be nice with a small test here.
That is super quick to implement a unit test for it.

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -70,43 +67,56 @@ const SearchPageLayout = ({ searchQuery }: { searchQuery?: string }) => {
...
+      // TODO: This should be configurable.
+      limit: 10,
+      filters: {
+        branchId: branchIds,
+        ...facetsForSearchRequest,
+      },
     },
-  })
+    {
+      enabled: q?.length > 0,
```
spaceo, line 87: Should be greater than 2 right?
And here we are back to some static configuration :)

spaceo (changes requested):
Nice job.
One thing needs to be clarified though (in two places):
https://github.com/danskernesdigitalebibliotek/dpl-go/pull/11/files/259b9970e0c1e03bff7259c47c3b07015c41b847..910199cc860276a6e272e2e881d0f04a3466208c#r1812486908

components/shared/searchFilters/helper.ts
Adamik10: image
It doesn't seem like that's an option. Is it a big deal though?

components/pages/searchPageLayout/SearchPageLayout.tsx
Adamik10: I'd say for now we can keep it more than 0. I'd say because the target audience is children they might not be able to easily understand that they need min 3 letters. And we can totally make queries with just 1 letter.

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -70,43 +67,56 @@ const SearchPageLayout = ({ searchQuery }: { searchQuery?: string }) => {
...
+      facetLimit: 100,
+      facets: facetDefinitions,
+      filters: {
+        branchId: branchIds,
+        ...facetsForSearchRequest,
+      },
     },
-  })
+    {
+      enabled: q?.length > 0,
```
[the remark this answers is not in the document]
Adamik10: I'd say because the target audience is children they might not be able to easily understand that they need min 3 letters. It was the project specs and Tue pushing for it to be 3 chars on the adult page. But here there is no project specs that micro manage us in the same way, and we can totally make queries with just 1 letter.

components/shared/searchFilters/helper.ts
spaceo: Nope :)

PR #12, opened 2024-10-22
Login sheet + user page stub

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-112

#### Description

This PR adds a login sheet for anonymous users to be able to login into Unilogin.
It also changes the user button in header to link to user profile page for users that are logged in.
The user profile page has a button for logging out again.
And there is session debugging showing info about the authorized user.

#### Notes

Sneaked in a temporary way of resolving config. Since we need it in the Lagoon PR environments.
We need to find out how to treat config...

#### Screenshot of the result

image
image


Adamik10 (changes requested):
Mikkel is there a separate task for the layout of the login sheet that comes from the side? 
I would say either make a ticket for it, or adjust the styling so that it matches:
- desktop: https://www.figma.com/design/AJs4sKPcAFdAnZlJZ5MkZz/eReolen-GO!---Designsystem-(Delt-med-kunden)?node-id=182-91150&t=8y9WPIk8KKQiWmvb-1
- mobile: https://www.figma.com/design/AJs4sKPcAFdAnZlJZ5MkZz/eReolen-GO!---Designsystem-(Delt-med-kunden)?node-id=182-92291&t=8y9WPIk8KKQiWmvb-1

lib/utils.ts
```diff
@@ -1,6 +1,37 @@
-import { type ClassValue, clsx } from "clsx";
-import { twMerge } from "tailwind-merge";
+import { extendedTheme } from "@/tailwind.config"
+import { type ClassValue, clsx } from "clsx"
+import { extendTailwindMerge } from "tailwind-merge"
+
+function formatColors() {
```
spaceo: I suggest: `formatTailwindColors` so we know the context/scope of the function.
And I suggest that both functions in this file goes into a file called:  `lib/helpers/tailwind.ts`

spaceo (commented):
@ThomasGross 
Please look at my comment..

lib/utils.ts
```diff
@@ -1,6 +1,19 @@
...
+  extend: {
+    classGroups: {
+      'font-size': [
+        {
+          text: Object.keys(extendedTheme.fontSize),
+        },
+      ],
+    },
+  },
+})
```
spaceo, line 15: I think we should introduce a namespace for global functionality like this. 
I suggest: `lib/helpers`
And this file could then be placed in there like: `lib/helpers/tailwind.ts`

spaceo (commented):
@thomas have one comment...

lib/utils.ts
Adamik10: I am down for `lib/helpers/` (I'm already using it in one of my PRs 😛

spaceo:
Thx for the feedback. I'll look into it.
But this one I dont understand why you report. Because it seems fine to me:
> 3. The actual LOGIN button inside the side-modal should be black in day-mode as can be seen in Figma [here](https://www.figma.com/design/AJs4sKPcAFdAnZlJZ5MkZz/eReolen-GO!---Designsystem-(Delt-med-kunden)?node-id=182-91150&t=7NHpt6SLKEL61V7x-1)
I have this:
image

spaceo:
@Adamik10 
Thomas and I looked at it.
We have fixed the border problem and the mobile width issue.
We have a thing about button backgrounds that are inconsitent in the Figma layout that we need to talk about.
Right now we have decided to make the bg white like everywhere else.
There is a thing about transition duration and the dark/light switch that needs to be handled so we create a separate task for that.

PR #13, opened 2024-10-25
feat: refactor ESLint and Prettier configurations

ThomasGross:
#### Link to issue

No issue 

#### Description

Removed sorting in eslint and moved it so its handled in prettier


Adamik10 (approved):
Amazing, showstopping, brilliant, talented, never seen before, never the same, always unique, afraid to be referenced or not referenced

PR #14, opened 2024-10-25
Ddfbra 128 cover service

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-128

#### Description
This PR introduces:
- orval code generation capabilities into the project
- makes covers on search result page dynamic

#### Screenshot of the result
-

Note that the covers CANNOT be loaded in a classic browser like Chrome, because of the CORS error. If you want to run this and for it to work, open a special instance of Google Chrome by running this command in your terminal:

`open -n -a /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --args --user-data-dir="/tmp/chrome_dev_test" --disable-web-security`


next.config.mjs
```diff
@@ -17,7 +17,7 @@ const nextConfig = {
     remotePatterns: [
       {
         protocol: "https",
-        hostname: "i.pinimg.com",
+        hostname: "res.cloudinary.com",
```
spaceo, line 20: Ah yes the covers are hosted on Cloudinary - I forgot about that.

lib/types/ids.ts
```diff
@@ -0,0 +1,17 @@
...
+  | "recommender"
+  | "something-similar"
+  | "favorites-list-mc"
+  | "inspiration-recommender"
+  | "recommended-material"
+  | "recommendation"
+  | "material-grid-automatic"
+  | "material-grid-manual"
+
+export type LoanId = number
```
spaceo, line 17: Great 👍

lib/graphql/fragments/work.fbi.graphql
```diff
@@ -17,4 +17,12 @@ fragment WorkTeaser on Work {
       code
     }
   }
+  manifestations {
+    all {
+      ...ManifestationCover
```
spaceo, line 22: I like the naming. It is clear 👍

lib/token.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param {TokenType} type
+ *
+ * @returns {string} token
+ */
+export function getToken(type) {
+  return tokens[type]
+}
+
+export default getToken
```
spaceo, line 54: We have decided to get the bearer token in the fetcher like we do in the FBI client.
That will deprecate the set/get token system that is added here.

lib/fetchers/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+          params.append(key, inner.toString())
+        })
+      } else {
+        params.append(key, value.toString())
+      }
+    })
+  }
+
+  return params
+}
```
spaceo: Is this something that maybe belongs to a general way of handling url parameters?
We do something similar in the search filters right.
Consider consolidating the code if possible.

lib/fetchers/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+  url,
+  params,
+}: {
+  baseUrl: string
+  url: string
+  params: unknown
+}) => {
+  const urlParams = params ? `?${buildParams(params as FetchParams)}` : ""
+  return `${baseUrl}${url}${urlParams}`
+}
```
spaceo: This is something that is needed mostly by the REST fetchers, right? Just wonder if we need to make it clear that this is a functionality that is relevant in that scope only and not in the graphql context because graphql is always POST'ing to the same url.
Maybe I am just rambling here, and you can go along 🤯

lib/fetchers/helper.ts
```diff
@@ -50,3 +50,9 @@ export const getServiceUrlWithParams = ({
   const urlParams = params ? `?${buildParams(params as FetchParams)}` : ""
   return `${baseUrl}${url}${urlParams}`
 }
+
+export const filterFalsyValuesFromArray = <T>(
+  array: (T | undefined | null | false | 0 | "")[]
+): T[] => {
+  return array.filter(Boolean) as T[]
+}
```
spaceo: I'd say that this is so global that you could consider moving it to a global place. Like `lib/helpers/arrays.ts`?
What do you think?

components/shared/workCard/WorkCard.tsx
```diff
@@ -2,32 +2,55 @@ import Image from "next/image"
...
 
 const WorkCard = ({ work }: WorkCardProps) => {
+  const { data } = useGetCoverCollection({
+    type: "pid",
+    identifiers: getAllWorkPids(work),
+    sizes: ["small", "large", "default", "original"],
+  })
+  const bestRepresentation = work.manifestations.bestRepresentation
+  const allPids = [bestRepresentation.pid, ...getAllWorkPids(work)]
+  const coverSrc = getCoverUrls(data, allPids || [], ["small", "large", "default", "original"])
```
spaceo: I think the cover sizes should be defined once. Either in a `const` before the `WorkCard` component but preferably in a global static variable (which there will be a system developed for in the near future).
I suggest a TODO message saying that cover sizes will be handled in a global configuration in the future

lib/helper.ts
```diff
@@ -0,0 +1,17 @@
...
+  "bg-accents-pink",
+  "bg-accents-purple",
+  "bg-accents-orange",
+  "bg-accents-blue",
+] as const
+
+export const getRandomContentColorClass = () => {
+  const randomIndex = Math.floor(Math.random() * contentColorClasses.length)
+  return contentColorClasses[randomIndex]
+}
```
spaceo, line 17: Nice with global color functionality. Consider moving the functions to a file like: `lib/helpers/colors.ts`

spaceo (changes requested):
Great job!
Really enjoyed the clear separation of the commits. MAde my life as a reviewer a breeze.

We talked about skipping the set/get token part and do it, for now, as we handle the bearer token in the fbi fetcher.
Otherwise please consider my suggestions of structuring the global helpers.

.eslintignore
```diff
@@ -0,0 +1,2 @@
+lib/cover-service-api/model/
+lib/cover-service-api/cover-service.ts
```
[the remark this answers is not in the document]
Adamik10: I think, because we use prettier to format code - the formatting is still done on "save". 
As far as I know, ESlint doesn't really format code in any case. It only points out when it's not living up to the standards.

tsconfig.json
```diff
@@ -28,5 +28,9 @@
     }
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
-  "exclude": ["node_modules"]
+  "exclude": [
+    "node_modules",
+    "lib/cover-service-api/model",
+    "lib/cover-service-api/cover-service.ts"
```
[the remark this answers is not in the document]
Adamik10: I believe prettier is taking care of the formatting. This is just so we're not getting TS errors in generated files.

lib/token.js
Adamik10: Roger! I deleted this file and am now following the way you guys get the token from the env file :)

lib/fetchers/helpers.ts
Adamik10: I feel like this is quite specifically targeted to creating search parameters for fetchers. Filter params we work with in a different way where we need to toggle them on and off. 
You're right that this function is a more generic one, but I don't think we can merge it / use it with filters.

lib/fetchers/helpers.ts
Adamik10: I adjusted the name so that it makes it more clear 😉

lib/fetchers/helper.ts
Adamik10: I did that in a later commit 😉

lib/helper.ts
Adamik10: I created a global helper/ directory with different helper files - `helper.colors.ts` is one of them 😉

PR #15, opened 2024-10-28
feat: add grid helper component and update global styles

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-152

#### Description

- Add a GridHelper component to improve workflow with the grid.
- Updated the global styles in globals.css to include new grid classes.

#### Screenshot of the result

Screenshot 2024-10-28 at 10 45 41


app/layout.tsx
```diff
@@ -36,6 +36,7 @@ export default function RootLayout({
   return (
     <html lang="en">
       <body className={`${GTFlexa.variable} antialiased`}>
+        <GridHelper />
```
spaceo, line 39: I think we need a check if we are in develop environment here. We do not want to include this in our hosting environments.

components/global/gridHelper/GridHelper.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    window.addEventListener('keydown', handleKeydown)
+    return () => {
+      window.removeEventListener('keydown', handleKeydown)
+    }
+  }, [])
+
+  const handleKeydown = (event: KeyboardEvent) => {
+    if (event.metaKey && event.key === 'g') {
```
spaceo, line 16: This could possibly be conflicting with cases where we need to use the letter g and then it would be annoying if the grid accidentally shows up, right?
Maybe we could have an optional environment variable called `NEXT_PUBLIC_GRID_DISABLED` and if that is set (to true) then it is not possible to trigger the grid?

spaceo (changes requested):
Great addition! 👏 
Please look at my comments. Most importantly the one about if the functionality should be used in our hosting env.

Adamik10 (commented):
I would like us to take this up altogether before it's merged. 
I have never worked with a helper like this outside of Figma when designing stuff, and I'm not even sure what the exact expectation for using this is. Let's talk about it at the daily tomorrow 😉

spaceo (approved):
LGTM 👍 
But lets wait with merging until we are aligned with how to work with grid layout in the team

PR #16, opened 2024-10-28
Ddfbra 129 blue titles

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-129

#### Description
This PR mainly focuses on:
- adding the "blue" indicator on search result cards if a work is costfree. 
- Introducing publizon adapter into orval.config

I also:
- prettified some files (we should talk about common tools that were introduced in the .vscode.example folder a coupld weeks ago)
- fixed some styling issues

#### Screenshot of the result
-

-


lib/helpers/ids.ts
ThomasGross, line 1: If this helper only is tied to publizon-api I would consider putting it inside a publizon-api helper

lib/publizon-api/mutator/fetcher.ts
ThomasGross, line 1: I think we have a fetchers folder consider putting this file there

tailwind.config.ts
```diff
@@ -283,7 +289,7 @@ const config: Config = {
     "./components/**/*.{js,ts,jsx,tsx,mdx}",
     "./app/**/*.{js,ts,jsx,tsx,mdx}",
     "./styles/**/*.{js,ts,jsx,tsx,mdx}",
-    "./lib/**/*.{js,ts,jsx,tsx,mdx}",
+    "./lib/**/*.{js,ts,jsx,tsx,mdx,colors}",
```
ThomasGross, line 292:
```suggestion
    "./lib/**/*.{js,ts,jsx,tsx,mdx}",
```

components/shared/badge/Badge.tsx
```diff
@@ -1,20 +1,18 @@
...
-          "border-transparent bg-primary text-primary-foreground shadow hover:bg-primary/80",
-        secondary:
-          "border-transparent bg-secondary text-secondary-foreground hover:bg-secondary/80",
-        destructive:
-          "border-transparent bg-destructive text-destructive-foreground shadow hover:bg-destructive/80",
-        outline: "text-foreground",
+        default: "bg-foreground text-background",
+        outline: "bg-secondary border-1 border border-foreground text-foreground",
+        destructive: "bg-destructive text-destructive-foreground",
+        blue: "bg-blue-title",
```
ThomasGross, line 15: Consider either calling the blue variable something not tied to blue titles or specific to blue titles

```suggestion
        blue: "bg-blue",
```
```suggestion
        "blue-title": "bg-blue-title",
```

components/shared/workCard/WorkCard.tsx
```diff
@@ -41,7 +57,12 @@ const WorkCard = ({ work }: WorkCardProps) => {
...
       <Link href={`/work/${work.workId}`}>
-        <div key={work.workId} className="rounded-sm bg-background-overlay p-2 md:p-4">
+        <div key={work.workId} className="relative rounded-sm bg-background-overlay p-2 md:p-4">
+          {!!dataPublizon?.product?.costFree && (
+            <Badge variant={"blue"} className="absolute left-2 md:left-4 md:top-4">
+              BLÅ
+            </Badge>
+          )}
           <div
             className="px-auto relative mx-auto mb-3 mt-6 flex aspect-[166/228] w-[calc(100%-76px-4px)] items-center
```
ThomasGross, line 67: - `px-auto` is not at valid class
- consider width and padding to be a bit more dynamic w-[80%]  
- I suggest removing aspect ratio and create a container where the image can be placed inside and use its own intrinsic image dimensions 

Something like this

```suggestion
            className="relative mx-auto mb-3 mt-6 flex w-[80%] items-center
```

ThomasGross (approved):
Very good work! 🙏

I have some minor things that maybe can be improved a bit.

lib/helpers/ids.ts
Adamik10: I kept it as is, because it's tied to all kinds of IDs between services

components/shared/workCard/helper.ts
```diff
@@ -1,11 +1,11 @@
 import { flatten } from "lodash"
 
-import { Cover } from "@/lib/cover-service-api/model"
 import { SearchWithPaginationQuery, WorkTeaserFragment } from "@/lib/graphql/generated/fbi/graphql"
 import { filterFalsyValuesFromArray } from "@/lib/helpers/arrays"
 
-import { CoverImageUrls } from "./../../../lib/cover-service-api/model/coverImageUrls"
 import { GeneralMaterialTypeCode } from "./../../../lib/graphql/generated/fbi/graphql"
+import { Cover } from "./../../../lib/rest/cover-service-api/generated/model"
+import { CoverImageUrls } from "./../../../lib/rest/cover-service-api/generated/model/coverImageUrls"
```
ThomasGross, line 8: I generally like imports routes to be set from project root. 

Something like this
```suggestion
import { CoverImageUrls } from "@/*??*/*??*/lib/rest/cover-service-api/generated/model/coverImageUrls"
```

It becomes easier to move files around and update imports if the import route is not rooted in the actual file. Consider doing this.

ThomasGross (approved):

components/shared/workCard/WorkCard.tsx
Adamik10: I removed the padding class :)))

Without the aspect, the cards will collapse depending on the cover, which doesn't look desireable:
image
image
I'm keeping the aspect ratio :3

components/shared/workCard/helper.ts
Adamik10: Wonderful suggestion!!! THX!!!

ThomasGross: Great i tend to sometimes forget myself. Please also correct me 😊

components/shared/workCard/WorkCard.tsx
ThomasGross: Okay didn't expect that to happen. Keep the aspect ratio and simplify it?  

Maybe try whatever is closest of the already defined aspect ratios? 
We have those in the config. If none fit then create a new one. 

 aspectRatio: {
    "1/1": "var(--aspect-1-1)",
    "1/2": "var(--aspect-1-2)",
    "2/1": "var(--aspect-2-1)",
    "2/3": "var(--aspect-2-3)",
    "3/2": "var(--aspect-3-2)",
    "3/4": "var(--aspect-3-4)",
    "4/3": "var(--aspect-4-3)",
    "4/5": "var(--aspect-4-5)",
    "5/2": "var(--aspect-5-2)",
    "5/4": "var(--aspect-5-4)",
    "16/9": "var(--aspect-16-9)",
  },

PR #17, opened 2024-10-29
Fix various styling inconsistencies

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-112

#### Description
This PR fixes styling connected to logging-in so that it work between day & night mode + is accessible.
I also changes "loading" into "isLoading" so that it makes sense that it's a boolean 😉 

#### Screenshot of the result
-

-


Dresse (approved):
LGTM 👍

PR #19, opened 2024-10-29
Ddfbra 92 paginering

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-92

#### Description

This PR contains pagination for the search result page. To make it work it was necessary to use multiple useEffect. This would be nice to clean up in the future.


app/search/fetchSearchResult.server.ts
```diff
@@ -6,6 +6,14 @@ import {
   useSearchWithPaginationQuery,
 } from "@/lib/graphql/generated/fbi/graphql"
 
+const facetTypes = [
+  "materialTypesGeneral",
+  "mainLanguages",
+  "age",
+  "lix",
+  "subjects",
+] as FacetField[]
```
spaceo: Great to have them defined only once. In the future (when the config setup is ready) it should be defined there, I think.

components/shared/workCard/WorkCard.tsx
```diff
@@ -4,39 +4,37 @@ import React from "react"
...
-              src="https://i.pinimg.com/564x/b0/4a/b2/b04ab2de27591aee4a76015e42e0e282.jpg"
-              alt="work image"
-              layout="responsive"
-              width={166}
-              height={228}
-              className="overflow-hidden rounded-sm"
-            />
-          </div>
-          <WorkCardAvailabilityRow materialTypes={work.materialTypes} />
+    <Link href={`/work/${work.workId}`}>
```
spaceo: I think we should have a global route helper function. We probably will need the work route in several places.
So it could eg. be called `resolveWorkUrl(work: Work)`

app/search/fetchSearchResult.server.ts
```diff
@@ -1,23 +1,23 @@
...
 } from "@/lib/graphql/generated/fbi/graphql"
-import { facetDefinitions } from "@/components/shared/searchFilters/helper"
 
 const prefetchSearchResult = async (q: string, queryClient: QueryClient) => {
   await queryClient.prefetchQuery({
     queryKey: useSearchWithPaginationQuery.getKey({
       q: { all: q },
       offset: 0,
-      // TODO: This should be configurable.
+      // TODO: This should match the query on search page
```
spaceo, line 14: ...and be configurable (but that will be solved in another task :))

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -45,7 +49,11 @@ const branchIds = [
   "775130",
 ] as string[]
 
-export const formatFacetTerms = (filters: { [key: string]: { [key: string]: FacetValue } }) => {
+const SEARCH_RESULTS_LIMIT = 9
```
spaceo, line 52: This is an example of how it can go wrong when we do not have a fixed places for configuration. Here it is 9 but the limit defined in `app/search/fetchSearchResult.server.ts` is 10.

spaceo: ...but the configuration part will be solved in the near future.

components/pages/searchPageLayout/SearchPageLayout.tsx
```diff
@@ -99,24 +126,73 @@ const SearchPageLayout = ({ searchQuery }: { searchQuery?: string }) => {
...
+      <h1 className="mt-[88px] text-typo-heading-2">{`Viser resultater for "${q}" ${data?.pages?.[0]?.search.hitcount ? "(" + data?.pages?.[0]?.search.hitcount + ")" : ""}`}</h1>
+      {/* TODO: add ghost loading and cleanup the code below  */}
       {isLoadingFacets && <p>isLoadingFacets...</p>}
       {!facetData?.length && <p>Ingen filter</p>}
       {facetData && facetData?.length > 0 && <SearchFilterBar facets={dataFacets.search.facets} />}
       {isLoading && <p>isLoading...</p>}
-      {searchData?.hitcount === 0 && <p>Ingen søgeresultat</p>}
-      {searchData?.works && <SearchResults works={searchData.works} />}
+
+      {data?.pages?.[0]?.search.hitcount === 0 && <p>Ingen søgeresultat</p>}
```
spaceo, line 179: This is mostly a matter of preference but I am not of repeating the same optional chaining. If we can resolve before the rendering if there are any `data.pages` it is more clean to me. Instead of asking: "do we have data.pages now and do we have dapa.pages now and..." :)

spaceo (approved):
I like 🎉 
I understand that you are not fund of the multiple useEffects and useStates.
Let's get this in. I would like to have it transformed into an xstate machine. Maybe I will look into that in the future.
But nice work 👍

app/search/fetchSearchResult.server.ts
ThomasGross: agree

components/shared/workCard/WorkCard.tsx
ThomasGross: I will create a ticket on this

PR #21, opened 2024-10-30
1. version of config handling

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-142

#### Description

This PR adds a concept of handling config wether it is coming from the environment, is statically defined config or is coming from a configuration endpoint.

The configuration endpoint part will be fully developed when the query is ready in the dpl-cms.

#### Screenshot of the result

image


Adamik10 (approved):
Great piece of work :)

PR #33, opened 2024-11-06
feat: add route resolver for work URLs and update WorkCard component

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-159

#### Description

Add a resolver function that streamline the internal routes in the app


Adamik10 (approved):
Wonderful!

PR #40, opened 2024-11-11
refactor: route helper function and add unit tests

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-159

#### Description

Refactored route helper function and add unit tests


lib/helpers/helper.routes.ts
```diff
@@ -1,34 +1,48 @@
...
 
-// Example usage:
-// const userRoute = buildRoute("/users/:id", { id: 123 })
-// console.log(userRoute) // Output: /users/123
-
-// const searchRoute = buildRoute("/search", undefined, { query: "test" })
-// console.log(searchRoute) // Output: /search?query=test
-
-function buildRoute(route: string, params?: RouteParams, query?: QueryParams): string {
+export function buildRoute(route: string, params?: RouteParams, query?: QueryParams): string {
```
spaceo, line 4: Consider putting the three parameters in an object, then it is easier to handle optional parameters and you don't need to specify `undefined` like here:
https://github.com/danskernesdigitalebibliotek/dpl-go/pull/40/files#diff-4e5726f5588ca037aa726de6816a68efcc47e8e4c64ea260599410ee23dcecb1R44

spaceo (approved):
Looks good. Nice to have a central function for this.
Approved, but please look at comments 👍

PR #41, opened 2024-11-11
Ddfbra 145 automatisk tilgaengeligheds tests

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-145

#### Description
This PR adds accessibility testing tools to storybook + a GH action that runs these on every push.
One of the addons that is worth advertizing here is the `eyesight problem simulator` in our stories:
image
..which we can use to simulate different diagnoses in regards to sight.

#### Screenshot of the result
-

-


.github/workflows/accessibility-test.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Build Storybook
+        run: yarn build-storybook --quiet
+
+      - name: Serve Storybook and run tests
+        run: |
+          export FORCE_COLOR=1
+          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
+            "npx http-server storybook-static --port 6006 --silent" \
+            "npx wait-on tcp:6006 && yarn test-storybook"
```
spaceo, line 42: You can set env variables (equivalent to export) in the `env` property of the task. Like this:
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/45a3d81ed826eec779660180e950a718f9c233f1/Taskfile.yml#L393-L406

spaceo (approved):
Looks good. 🎉   I have one comment about using environment variables in task cmds...

PR #42, opened 2024-11-11
Use absolute  urls in unilogin flow redirects

spaceo:
`Response.redirect` needs absolute url in order to function correctly


Adamik10 (approved):
🥇

PR #44, opened 2024-11-12
Trying new way of using server source maps

spaceo:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-217

#### Description

Trying new way of using server source maps
Following this article:
https://stackoverflow.com/a/78111855


ThomasGross (approved):
LGTM

PR #48, opened 2024-11-15
feat: add UniLogin configuration query and update related services

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-97

#### Description

Add UniLogin configuration query and update related services


spaceo (approved):
LGTM 👍

PR #50, opened 2024-11-19
Ddfbra 200 anonym og indlogget bruger kan prove e bog

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-200

#### Description

This PR introduces GraphQL queries and fragments for material and work + update TypeScript configuration


Adamik10 (approved):
Let's get this in so we can both work furher. 
It would be worth investigating the types in the Reader component in the upcoming PR 😉

PR #58, opened 2024-11-25
Ddfbra 201 go anonym og indlogget bruger kan prove lydbog

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-201

#### Description

This PR includes 
- A publizon player module. 
- Shared dialog and drawer component


__tests__/url.test.ts
```diff
@@ -3,20 +3,19 @@ import { expect, test } from "vitest"
...
-    routeParams: { wid: 123 },
+    routeParams: { work: "work", wid: 123 },
     queryParams: { audio: "true" },
   })
   expect(workUrl).toBe("/work/123?audio=true")
 })
 
 test("That resolveUrl can return a search url", async () => {
-  const workUrl = resolveUrl({ type: "search", queryParams: { q: "test" } })
+  const workUrl = resolveUrl({ routeParams: { search: "search" }, queryParams: { q: "test" } })
```
spaceo, line 19: We talked about taking a look at the `resolveUrl` :) Let's do that on a rainy day.

components/pages/workPageLayout/WorkPageLayout.tsx
```diff
@@ -1,19 +1,57 @@
...
+
+  const { data } = useQuery({
     queryKey: useGetMaterialQuery.getKey({ wid }),
     queryFn: useGetMaterialQuery.fetcher({ wid }),
   })
 
+  const manifestations = data?.work?.manifestations.all
+
+  const identifier = manifestations?.[0].identifiers?.[0].value || ""
+
```
spaceo, line 24: What happens if we do not get data or work and therefore do not have manifestations? How is the rest of the code reacting?

components/shared/publizonPlayer/helper.ts
```diff
@@ -0,0 +1,29 @@
...
+    document.head.appendChild(scriptElement)
+  }
+}
+
+export const removeAsset = ({ src, type }: TAssetType) => {
+  if (type === "script") {
+    const scriptElement = document.querySelector(`script[src="${src}"]`)
+    scriptElement?.remove()
+  }
+}
```
spaceo, line 29: Why having two sets of `appendAsset` `removeAsset` functions?
As far as I see we can a common way of doing this in `/lib/helpers/reader-player/...` and share it among the components.
You could consider making your custom hook adding the assets receiving an array of assets - which you could use in both the reader and player?

spaceo (approved):
I think you could consider consolidating functionality that is repeated in both the reader and player and define it once.
Othjerwise, looks good 👍

components/pages/workPageLayout/WorkPageLayout.tsx
ThomasGross: TODO is added here

components/shared/publizonPlayer/helper.ts
ThomasGross: Great idea. I will do that and merge the branch into main afterwards.

PR #60, opened 2024-11-27
Add some error handling in the dpl cms fetcher

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-226

#### Description

Add some error handling in the dpl cms fetcher
Maybe it is too radical. We need a general error handling strategy.
Also updated the .env.example with ned https dpl cms graphql endpoint.


spaceo:
> I think it's definitely worth talking about how we would like to handle unsuccessful requests. Does this just block the whole app if a request to the CMS is unsuccessful?
> 
> Consider how that affect the experience at the users' end. Would we maybe prefer to just not render certain components, or even create "couldn't load" versions of components like we have skeleton versions of them - in order to preserve site interactivity in areas that are not affected by an unsuccessful request?

My changes we introduced because we did not have any handling of errors and I could end in some error situations where it was difficult as an develop to understand why things were breaking.
I have reverted the changes for now. Because I don't have time for the full fledge UI handling of the errors that you mention.
I am curios how would you handle the problems that you address. Do you have any suggestions?

Adamik10 (approved):
💯

PR #61, opened 2024-11-27
feat: migrate Accordion component to shared directory

ThomasGross:
#### Link to issue

none

#### Description

This PR moves the accordion to the correct folder


Adamik10 (changes requested):
Omg so sorry that this was hanging for so long. Would you mind making a new PR out of main? I think there's been too many changes since, and I wouldn't want to lose the changes we did in the accordion component since two weeks ago.

Adamik10 (approved):
Live, laugh, love

PR #63, opened 2024-11-28
Xstate fixes

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-130

#### Description

This PR adresses the behaviour that changed after the transition to use Xstate in searching/filtering.
Things being handled:
* The skeleton screen on load more
* The behaviour around deleting last character in the search field (which was actually working as intended, but I agree in an oddly manner 😆 )
* The weird issue about FBI sending one kind of facet names from service and expecting another as input (lix -> lixRange)
* Changing the way filters where toggled and detected as being active (that was not handled after the inital xstate introduction)
* Allowing the end user to type and filter while requests are ongoing

### DIFF
--- a/__tests__/search.test.ts
+++ b/__tests__/search.test.ts
@@ -85,31 +85,6 @@ describe("Facet functionality", () => {
     expect(result).toStrictEqual(facetFilters)
   })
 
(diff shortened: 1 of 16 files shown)


ThomasGross (approved):
Looks good 🥳

PR #64, opened 2024-12-02
Add functionality for resolving dpl-cms graphql endpoint

spaceo:
#### Description
We do not have a way in Lagoon for resolving the url to the dpl-cms graphql endpoint.
This PR should fix that 🤞 

At least the tests are passing...


ThomasGross (approved):
LGTM

PR #67, opened 2024-12-03
Add "Blue title" badge for cost free works on work page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-173
https://reload.atlassian.net/browse/DDFBRA-248

#### Description
This PR:
- adds the blue title badge above the work title on work page for works that are cost free
- improves work title breaking into multiple lines based on the language ISO code that we can get from FBI API
- fixes minor bugs
- adds selected manifestation type to the URL on material page

#### Screenshot of the result
-

-


components/pages/workPageLayout/WorkPageHeader.tsx
```diff
@@ -51,12 +71,30 @@ const WorkPageHeader = ({ work }: WorkPageHeaderProps) => {
...
+  const languageIsoCode = getManifestationLanguageIsoCode(selectedManifestation)
+  const onOptionSelect = (optionSelected: SlideSelectOption) => {
+    setSelectedManifestation(
+      getManifestationByMaterialType(work, optionSelected.value) ||
+        work.manifestations.bestRepresentation
+    )
+    const searchParams = new URLSearchParams(window.location.search)
+    searchParams.set("type", optionSelected.render)
+    router.push(`${window.location.pathname}?${searchParams.toString()}`)
+  }
```
ThomasGross, line 97: Instead of setting the selected manifestation here. I would suggest setting the url param and make a useeffect that listens to changes on url param "type". Inside that useEffect set the new selected manifestation type. :) 

On route push function consider adding the following to in options
`{
      scroll: false
 }`
This will prevent scrolling to the top of the page when changing type.
https://nextjs.org/docs/app/api-reference/components/link#reference

ThomasGross (changes requested):
Looks good. I have some comments.

Jeg tried running the code and the switch seems to be working in the opposite way as intended. (when click on the active manifestation it switches) 

Also It could be nice to show the same icons on the switch as on the cards on the search result page. 😊

components/pages/workPageLayout/WorkPageHeader.tsx
Adamik10: It turned out to be quite an undertaking haha. But I did it <3

ThomasGross: Great! Looks good. Please add the icons to the switch or create a ticket on the issue. 

Screenshot 2024-12-05 at 14 49 26

ThomasGross (approved):
Looks good

PR #69, opened 2024-12-05
feat: add animation for header switch and remove global transition for now

ThomasGross:
#### Link to issue

No issue 

#### Description

- Removed global transition og light/dark mode switch
- Added animation to the switch in the header


Adamik10 (changes requested):
PLS PLS PLS PLS let's not remove the transition 🙏 
We can fix the "hacky code" we talked about and it still works 😉

Adamik10 (approved):
Yay! Thx 🎉

PR #70, opened 2024-12-06
Ddfbra 246 fix missing search results

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-246

#### Description

fix: bug that returns an empty search result when navigation with native browser navigation buttons
fix: reset offset when making new search and filtering
feat: add initial context as an globally accessible reference


Adamik10 (approved):
👏

PR #71, opened 2024-12-06
Move the Lagoon url resolving to the application build step

spaceo:
#### Description

Move the Lagoon url resolving to the application build step.
Because the LAGOON_ROUTES env variable is not accessible client side, we do the url processing before the build step and turn them into NEXT_PUBLIC_ variables.


Adamik10 (approved):
LGTM :)

PR #72, opened 2024-12-09
fix: show ghost loading both when loading more results and initial results

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-246

#### Description

Look in the title


Adamik10 (approved):
Gorge

PR #79, opened 2024-12-17
Show work details info

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-170

#### Description
This PR:
- adds the InfoBoxDetails component 
- adjusts some styling 
- fixes conditional rendering of work description items
- adds storied for InfoBox & InfoBoxDetails components

#### Screenshot of the result
image
image

I can't figure out why the InfoBox stories are full width and InfoBoxDetails are not, as they're styled in the same way. Pls if anyone knows, do tell :))

### DIFF
--- a/components/shared/infoBox/InfoBox.tsx
+++ b/components/shared/infoBox/InfoBox.tsx
@@ -28,29 +28,20 @@ const InfoBox = ({ work }: InfoBoxProps) => {
           </div>
           <dl className="flex-1">
             {!!selectedManifestation?.audience?.ages.length && (
-              <>
-                <hr className="border-border not-first:mt-7" />
-                <InfoBoxItem
-                  term="Alder"
-                  description={selectedManifestation.audience.ages.map(age => age.display)}
-                />
-              </>
+              <InfoBoxItem
+                term="Alder"
+                description={selectedManifestation.audience.ages.map(age => age.display)}
+              />
             )}
             {!!selectedManifestation?.series?.length && (
-              <>
-                <hr className="border-border not-first:mt-7" />
-                <InfoBoxItem term="Serie" description={getSeriesInfo(selectedManifestation)} />
-              </>
+              <InfoBoxItem term="Serie" description={getSeriesInfo(selectedManifestation)} />
             )}
             {!!selectedManifestation?.subjects.all?.length && (
-              <>
-                <hr className="border-border not-first:mt-7" />
-                <InfoBoxItem
-                  term="Emneord"
-                  description={selectedManifestation.subjects.all.map(subject => subject.display)}
-                  isButtons
-                />
-              </>
+              <InfoBoxItem
+                term="Emneord"
+                description={selectedManifestation.subjects.all.map(subject => subject.display)}
+                isButtons
+              />
             )}
           </dl>
         </div>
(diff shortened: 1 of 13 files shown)


components/shared/infoBox/InfoBox.tsx
```diff
@@ -19,39 +19,30 @@ const InfoBox = ({ work }: InfoBoxProps) => {
...
-            {work.abstract &&
-              work.abstract.length &&
-              work.abstract.map(abstract => (
-                <p key={abstract.substring(0, 30)} className="pt-5 lg:pt-0">
-                  {abstract}
-                </p>
-              ))}
+            {!work.abstract?.length && <p>Værket har desværre ingen beskrivelse.</p>}
+            {!!work.abstract?.length &&
+              work.abstract.map(abstract => <p key={abstract.substring(0, 30)}>{abstract}</p>)}
```
JacobArrow, line 27: This would be more readable

```suggestion
{!work.abstract?.length ? (
  <p>Værket har desværre ingen beskrivelse.</p>
) : (
  work.abstract.map(abstract => (
    <p key={abstract.substring(0, 30)}>
      {abstract}
    </p>
  ))
)}
```

components/shared/infoBox/InfoBox.tsx
```diff
@@ -19,39 +19,30 @@ const InfoBox = ({ work }: InfoBoxProps) => {
...
-                  isButtons
-                />
-              )}
+            {!!selectedManifestation?.subjects.all?.length && (
+              <InfoBoxItem
+                term="Emneord"
+                description={selectedManifestation.subjects.all.map(subject => subject.display)}
+                isButtons
+              />
+            )}
```
JacobArrow, line 45: Have you considered if we want any fallback, if theese conditions are not met?

components/shared/infoBox/InfoBoxDetails.tsx
JacobArrow, line 1: This file seems very cluttered with all the conditionals, is there really no better data integrity?
Do we need fallbacks if conditionals fail?

components/shared/infoBox/InfoBoxItem.tsx
```diff
@@ -20,17 +20,17 @@ const InfoBoxItem = ({ term, description, isButtons = false }: InfoBoxItemProps)
...
       {isButtons && (
-        <div className="mt-6 flex flex-row flex-wrap gap-2">
+        <dd className="mt-6 flex flex-row flex-wrap gap-2">
           {description.map(item => (
             <Button key={item} size={"sm"} className="px-3" onClick={() => handleClick(item)}>
               {item}
             </Button>
           ))}
-        </div>
+        </dd>
```
JacobArrow, line 33: Good catch with the semantics here!

JacobArrow (approved):
Great work on this!

components/shared/infoBox/InfoBox.tsx
Adamik10: perfecto

components/shared/infoBox/InfoBox.tsx
Adamik10: Yesss, we agreed with them to not show the items then.

components/shared/infoBox/InfoBoxDetails.tsx
Adamik10: I understand what you mean. I still stand by this though if you don't have any better ideas. 
And no need for fallbacks I'd say. We agreed with them to not show the data pairs if there's no data for each item. 
Maybe one day if we see that there's often a lot of data missing could consider rendering dashes or something so that the rows don't just not appear, but for now we trust that there are no big holes in the data.

PR #81, opened 2024-12-19
Material types in singular

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-253

#### Description
This PR:
- shows material types in SlideSelect in singular
- shows material types in Details info box in singular
- shows a dash "-" for empty values for improved UX
- shows contributors in work page header in case work creators aren't defined

#### Screenshot of the result
image
image
image

-


components/shared/workCard/helper.ts
ThomasGross, line 1: displayCreators method should be moved to a global helper file now its not only used in workcard

components/shared/workCard/helper.ts
```diff
@@ -1,19 +1,22 @@
...
   GeneralMaterialTypeCodeEnum,
+  ManifestationDetailsFragment,
   SearchWithPaginationQuery,
+  WorkFullWorkPageFragment,
   WorkTeaserSearchPageFragment,
 } from "@/lib/graphql/generated/fbi/graphql"
 
 export const displayCreators = (
-  creators: SearchWithPaginationQuery["search"]["works"][0]["creators"],
+  creators: WorkFullWorkPageFragment["creators"] | ManifestationDetailsFragment["contributors"],
```
ThomasGross, line 10: I'm suggesting not having the function to be specific to the types WorkFullWorkPageFragment["creators"] and  ManifestationDetailsFragment["contributors"]. Could it expect an array instead here?

```suggestion
  creators: string[],
```

ThomasGross (approved):
Looks good i have a few comments to consider

components/shared/workCard/helper.ts
Adamik10: That's not specific enough for this case. Then TS wouldn't be happy with us giving it an array of objects. I think it's pretty nice for a specific use case :)

PR #84, opened 2025-01-06
Introduce Cypress

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-233

#### Description

This PR adds the e2e framework Cypress to the project and includes a basic test that verify that the application builds successfully and that the frontpage renders a header and a footer. 

Future work will include mocking FBI data and testing core user journeys

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@
   Drupal CMS can be accessed through Lagoon by generating a one-time login in the preferred environment.
   <br>
   <br>
-  Last updated: 27/11/2024
 </p>
 
 ## URLs
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ The application is now running at [https://localhost:3000](https://localhost:300
 - [Tailwind CSS][tailwind]
 - [ESLint][eslint] & [Prettier][prettier]
 - [Storybook][storybook] & [Chromatic][chromatic]
+- [cypress][cypress]
 - [Vitest][vitest]
 
 ### Project structure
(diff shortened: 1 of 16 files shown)


README.md
ThomasGross, line 1: At the top of the readme there is a 'last updated'. Can you update it or maybe remove it? What do you think?

components/global/header/Header.tsx
```diff
@@ -5,17 +5,18 @@ import DarkModeToggle from "@/components/shared/darkModeToggle/DarkModeToggle"
...
 function Header() {
   return (
-    <div>
+    <header>
       <div className="flex h-navigation-top-height items-center justify-center bg-background-overlay">
         <p className="text-typo-caption">Biblioterernes ebøger og lydbøger</p>
       </div>
       <div className="content-container grid h-navigation-height grid-cols-3 items-center">
-        <div className="flex-0 flex items-center">
+        <div className="flex-0 flex items-center" data-cy={cyKeys["go-logo"]}>
```
ThomasGross, line 19: I can't figure out what I think about adding code to elements that only have relevance for Cypress. I know its a common practice but still worth mentioning?

I would like our components to be as clean as possible, preferably without extra data attributes.

If we keep it, I think either `data-cy={cyKeys["go-logo"]}` should be placed on `Button` or `Icon`"

ThomasGross (approved):
Great job, I have a few questions but overall approved ✅

components/global/header/Header.tsx
JacobArrow: It is the best practice for selecting elements with Cypress and I think we should stick with that :)
https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements

It also allows to typecheck our element selectors, which is always nice.

Regarding where it is placed; it doesnt matter in this PR, as it is only for demonstrating the usage. 
`frontpage.cy.ts` will be updated once we actually have some content to test.

PR #91, opened 2025-01-16
Codegen: Use unions instead of enums

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-234

#### Description

As a part of my work on Cypress and type safety, I discovered that it would make sense to use string unions instead of enums, as they can make our code more concise and readable.

### DIFF
--- a/__tests__/search.test.ts
+++ b/__tests__/search.test.ts
@@ -14,18 +14,25 @@ vi.mock(import("@/lib/config/goConfig"), async importOriginal => {
   }
 })
 
-export const facets = {
-  [FacetFieldEnum.Materialtypesgeneral]: {
+type Facets = {
+  [key in FacetFieldEnum]?: {
+    filter: string
+    translation: string
+  }
+}
+
+export const facets: Facets = {
+  MATERIALTYPESGENERAL: {
     filter: "materialTypesGeneral",
     translation: "Type",
   },
-  [FacetFieldEnum.Mainlanguages]: {
+  MAINLANGUAGES: {
     filter: "mainLanguages",
     translation: "Sprog",
   },
-  [FacetFieldEnum.Age]: { filter: "age", translation: "Alder" },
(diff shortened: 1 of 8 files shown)


__tests__/search.test.ts
```diff
@@ -14,18 +14,25 @@ vi.mock(import("@/lib/config/goConfig"), async importOriginal => {
   }
 })
 
-export const facets = {
-  [FacetFieldEnum.Materialtypesgeneral]: {
+type Facets = {
+  [key in FacetFieldEnum]?: {
```
spaceo, line 18: I thought that a part of this PR was to make codegen use `as const` instead of enums. I know that the Graphql still have the native type `enum` but then it could be converted?

components/shared/workCard/helper.ts
```diff
@@ -4,23 +4,18 @@ import {
...
+  gaming: GeneralMaterialTypeCodeEnum[]
+  video: GeneralMaterialTypeCodeEnum[]
+}
+
+export const materialTypeCategories: MaterialTypeCategories = {
+  reading: ["ARTICLES", "BOOKS", "COMICS", "EBOOKS", "IMAGE_MATERIALS", "NEWSPAPER_JOURNALS"],
+  listening: ["AUDIO_BOOKS", "MUSIC", "PODCASTS", "SHEET_MUSIC"],
+  gaming: ["BOARD_GAMES", "COMPUTER_GAMES"],
+  video: ["FILMS", "TV_SERIES"],
 }
```
spaceo, line 19: This ought to be converted into config in our config system. It has flown under my radar. :)

spaceo (approved):
Approved, but please look at my comments.

__tests__/search.test.ts
JacobArrow: The new generated enums will be a type of union like this: 

```
export type AccessTypeCodeEnum =
  | 'ONLINE'
  | 'PHYSICAL'
  | 'UNKNOWN';
```

PR #92, opened 2025-01-17
Ddfbra 334 podcast er altid bla titler

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-334
https://reload.atlassian.net/browse/DDFBRA-366
https://reload.atlassian.net/browse/DDFBRA-369

#### Description

- This PR makes all marks all podcasts material in searchresult and workpage as blue titles
- It highlights the blue title of a given manifestation in searchresult view
- Rewrite material


.env.example
```diff
@@ -1,6 +1,7 @@
 # YOUR DPL-CMS NEEDS TO BE RUNNING LOCALLY
 NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS=https://dapple-cms.docker/graphql
-NEXT_PUBLIC_GRAPHQL_BASIC_TOKEN_DPL_CMS=Z3JhcGhxbF9jb25zdW1lcjp0ZXN0
+# BASIC AUTH HEADER - YOUR LOCAL CONSUMER KEY AND SECRET FOR DPL-CMS
+NEXT_PUBLIC_GRAPHQL_BASIC_TOKEN_DPL_CMS="Z29fZ3JhcGhxbDpvQThtcVpGeG1GdWJSRUg3N1QzcWVBcGVvVXhvTlM3WQ=="
```
spaceo, line 4: Nice. Please create separate PR's for stuff like this in the future. 👍  Because then we can get it into main faster.

components/pages/workPageLayout/WorkPageHeader.tsx
```diff
@@ -1,51 +1,54 @@
...
-    return { value: materialType.code, render: materialType.display }
-  })
+  const workMaterialTypesWithDisplayName = workMaterialTypes.map(materialType => {
+    return {
+      code: materialType.code,
+      display: translateMaterialTypesStringForRender(
+        materialType.code as GeneralMaterialTypeCodeEnum
+      ),
+    }
+  }) as SlideSelectOption[]
```
spaceo, line 51: This is something to consider:
When you write a function inside the component (which essentially is another function) it will be **defined** every time the component is called (which can be quite a lot of times). It doesn't necessarily it will have an impact, but it can have an impact both in computation and in memory use.
Because of that I think it is a good practise, at least to consider writing the function outside of the component to avoid such pitfalls. Again it depends on the case but if the argument is: but this is easier, I don't care about the functions being pure it is definitely the wrong approach. Another benefit of having isolated functions that you can test them isolated i unit tests. That is not possible if they are defined inside the components.

spaceo: This also looks like a function that would be nice to have globally. But I could be wrong.

components/pages/workPageLayout/helper.ts
```diff
@@ -96,6 +69,24 @@ export const getManifestationLanguageIsoCode = (
...
+  "COMICS",
+  "COMPUTER_GAMES",
+  "FILMS",
+  "IMAGE_MATERIALS",
+  "MUSIC",
+  "NEWSPAPER_JOURNALS",
+  "OTHER",
+  "SHEET_MUSIC",
+  "TV_SERIES",
+]
```
spaceo, line 88: This, along with ..., should be moved to a config resolver. Like our [search config](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/031bd8badc536f034d77931604224563aed1ef8f/lib/config/resolvers/search.ts#L41-L54).

components/shared/infoBox/InfoBox.tsx
```diff
@@ -1,17 +1,39 @@
...
+    router.push(url, {
+      scroll: true,
+    })
+  }
+
+  const ageString = selectedManifestation?.audience?.ages.map(age => age.display).join(", ") || "-"
+  const seriesString =
+    selectedManifestation?.series
+      .map(series => `${series.numberInSeries ? series.numberInSeries + " i " : ""}${series.title}`)
+      .join(", ") || "-"
```
spaceo, line 33: I think these functions are so generic that they should be moved globally. Of course only if they are going to be used multiple places. But aren't they?
This is also functionality that we are doing again that we also have done in dpl-react, but I guess that we have already crossed that bridge.

components/shared/workCard/WorkCard.tsx
```diff
@@ -1,42 +1,92 @@
...
+              item.manifestation.materialTypes[0].materialTypeGeneral.code !==
+              current.manifestation.materialTypes[0].materialTypeGeneral.code
+          )
+          acc.push(current)
+        }
+      }
+      return acc
+    },
+    [] as typeof manifestationsWithPublizonData
+  )
```
spaceo, line 88: This is clearly a function that we should isolate )like I wrote earlier) because then we could test it in Vitest. I think you should laso write a test if you are up for it.

components/shared/workCard/helper.ts
```diff
@@ -9,13 +9,15 @@ type MaterialTypeCategories = {
...
 }
 
 export const materialTypeCategories: MaterialTypeCategories = {
   reading: ["ARTICLES", "BOOKS", "COMICS", "EBOOKS", "IMAGE_MATERIALS", "NEWSPAPER_JOURNALS"],
-  listening: ["AUDIO_BOOKS", "MUSIC", "PODCASTS", "SHEET_MUSIC"],
+  listening: ["AUDIO_BOOKS", "MUSIC", "SHEET_MUSIC"],
   gaming: ["BOARD_GAMES", "COMPUTER_GAMES"],
   video: ["FILMS", "TV_SERIES"],
+  podcast: ["PODCASTS"],
 }
```
spaceo, line 21: Should also be a conf resolver. The reason why I think that is also that we the can unify the materialtypes and categories in one place.

components/shared/infoBox/InfoBoxDetails.tsx
```diff
@@ -1,19 +1,25 @@
...
 type InfoBoxDetailsProps = {
-  work: WorkFullWorkPageFragment
+  selectedManifestation: ManifestationWorkPageFragment
 }
 
-const InfoBoxDetails = ({ work }: InfoBoxDetailsProps) => {
-  const { selectedManifestation } = useSelectedManifestationStore()
+const InfoBoxDetails = ({ selectedManifestation }: InfoBoxDetailsProps) => {
+  // get selectedManifestation materialTypes and translate them for render
+  const materialTypeDisplays = selectedManifestation?.materialTypes.map(materialType => {
```
spaceo, line 18: Please remove `?` from the `selectedManifestation?.*` entries since we know that `selectedManifestation` is always defined.

spaceo (commented):
Please look at my comments. This is not like it is **not**-approved but I would like you to evaluate my suggestions.

components/shared/workCard/WorkCard.tsx
ThomasGross: Im writing a TODO

PR #93, opened 2025-01-20
Redirect to search on click of author

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-171

#### Description
This PR adds functionality to redirect to search results with a search for the author upon clicking on their name on work page. As a part of this PR a new micro-animation is added in animate-text-underline. It can be used for text that links places. The design also matches FB CMS underlined text animation.

#### Screenshot of the result
https://github.com/user-attachments/assets/48a4683b-9157-48ed-9156-76811c7eb6d0

-


components/pages/workPageLayout/WorkPageHeader.tsx
```diff
@@ -150,9 +155,24 @@ const WorkPageHeader = ({ work }: WorkPageHeaderProps) => {
...
+                return (
+                  <>
+                    <span
+                      key={creator}
+                      onClick={() => navigateToSearch(creator, actor, router)}
+                      className="animate-text-underline">
+                      {creator}
+                    </span>
+                    {index + 1 < workCreators.length && ", "}
+                  </>
```
spaceo, line 171: Consider making two components: `WorkAuthor` and `WorkAthours` this is something that we easily could have the need for in the future. And I think it improves the readability.

components/shared/searchInput/helper.search.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const navigateToSearch = (
+  q: string,
+  actor: Actor<AnyActorLogic>,
+  router: AppRouterInstance
+) => {
+  if (!q) return
+  actor.send({ type: "SEARCH" })
+  router.push(`/search?q=${q}`, {
+    scroll: false,
```
spaceo, line 12: Just a note: @ThomasGross Made some kind of url resolving function I think we should use when we are dealing with urls. I think the resolving function could be improved but it would be noce that we handle urls in a uniform way.

lib/helpers/helper.creators.ts
```diff
@@ -19,3 +19,15 @@ export const displayCreators = (
...
+  creators: WorkFullWorkPageFragment["creators"] | ManifestationDetailsFragment["contributors"]
+) => {
+  return creators.reduce((acc, creator) => {
+    // Don't show one person twice
+    if (acc.includes(creator.display)) {
+      return acc
+    }
+    return [...acc, creator.display]
+  }, [] as string[])
+}
```
spaceo, line 33:
```suggestion
export const getAllCreators = (
  creators: WorkFullWorkPageFragment["creators"] | ManifestationDetailsFragment["contributors"]
) => [...new Set(creators.map(creator => creator.display))]
```

spaceo (changes requested):
Great addition. I like the animation.
Please look at my comment.
One thing that we need to talk about is about the search url.

components/pages/workPageLayout/WorkPageHeader.tsx
Adamik10: I like that! I think two components are maybe a bit of an overkill, but I made a component called `Authors.tsx` - without the "Work" prefix because it can render creators of both a work or a manifestation.

ThomasGross:
I like the animation on links. But we're introducing a new design concept and its should cascade through all similar items. My point is all similar links should have this animation or none. 

I feel like the figma design has a different concept for links in the footer and other places. I suggest we make sure that this is aligned 😊

tailwind.config.ts
```diff
@@ -347,6 +347,23 @@ const config: Config = {
...
+          bottom: "-5px",
+          left: "0",
+          width: "0",
+          height: "1px",
+          backgroundColor: "var(--foreground)",
+          transition: "width 0.3s ease",
+        },
+        ".animate-text-underline:hover::after": {
+          width: "100%",
+        },
```
ThomasGross, line 366: put these in the `@layer utilities` instead

components/shared/searchInput/helper.search.ts
ThomasGross: In my opinion, this change creates more confusion than it cleans up the code. :) 

Question for the team: I feel like these helpers in general is a bit overused and creates confusion. I like these helpers when the functions inside is used multiple places, but often they're not. I feel that we should only use helpers when needed and not as a default.

Adamik10: I just found that function after the changes have been merged! Nice catch :)

spaceo (approved):
Splendissimo 👍

PR #94, opened 2025-01-20
Ddfbra 294 hent paragraph data fra cms og vis i go fe

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-278

#### Description

This PR introduces:
- The video paragraph and related stories
- Typesafe way of dynamically rendering paragraphs 
- Go page rendering


app/[...slug]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const data = await fetcher<GetPageByPathQuery, { path: string }>(GetPageByPathDocument, {
+    path: slug.join("/"),
+  })()
+
+  const routeType = data.route?.__typename
+
+  if (routeType === "RouteRedirect") {
+    // TODO: implement redirect
+    return
```
Comment (spaceo) [inline app/[...slug]/page.tsx:25]: What happens when you return nothing?

components/pages/basicPageLayout/BasicPageLayout.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { NodeGoPage } from "@/lib/graphql/generated/dpl-cms/graphql"
+
+function BasicPageLayout({ pageData }: { pageData: NodeGoPage }) {
+  const { paragraphs } = pageData
+
+  return (
+    <div className="flex flex-col gap-y-paragraph-spacing">
+      <ParagraphResolver paragraphs={paragraphs ?? []} />
+    </div>
+  )
```
spaceo, line 13: I think that if we beforehand know that there is no data, in this case paragraphs I think we should not to to resolve pragraphs.
So it would be:
```typescript
  if (!paragraphs) {
    // Null or something else default.
    return nulll;
 }
  return (
    <div className="flex flex-col gap-y-paragraph-spacing">
      <ParagraphResolver paragraphs={paragraphs} />
    </div>
  )
```

components/paragraphs/ParagraphGoVideo/ParagraphGoVideo.tsx
```diff
@@ -0,0 +1,41 @@
...
+type ParagraphGoVideoProps = {
+  title: TParagraphGoVideo["title"]
+  url: TParagraphGoVideo["url"]
+}
+
+export default function ParagraphGoVideo(paragraphGoVideo: ParagraphGoVideoProps) {
+  const [mounted, setMounted] = React.useState(false)
+
+  useEffect(() => {
+    setMounted(true)
```
spaceo, line 16: Why do we need this? Please document why you need to keep in a state if the compoment is mounted.

app/[...slug]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (routeType === "RouteExternal") {
+    // TODO: implement external route redirect
+    return
+  }
```
Comment (spaceo) [inline app/[...slug]/page.tsx:31]: In general I am not fan of code stubs if they are not needed. Could we maybe just write one todo where we write that we need to take care of these two cases?
Consider creating a Jira task for it too.

codegen.ts
```diff
@@ -6,6 +6,7 @@ const { loadEnvConfig } = require("@next/env")
 
 loadEnvConfig(process.cwd())
 
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
```
spaceo, line 9: Please write a comment why this is needed.

components/pages/basicPageLayout/BasicPageLayout.tsx
spaceo: What kind of rendering do we want if there are no paragraphs?

components/paragraphs/ParagraphResolver.tsx
```diff
@@ -0,0 +1,31 @@
...
+      return null
+    }
+
+    const DynamicComponentType = components[type as keyof typeof components] || null
+    if (DynamicComponentType === null) return null
+
+    return (
+      <ParagraphErrorBoundary key={index}>
+        {/* @ts-ignore TODO: figure out how to type dynamically imported components */}
+        <DynamicComponentType key={index} {...paragraph} />
```
spaceo, line 25: While I think it is ok, for now, to  ignore the ts error,  I really think we should be able to handle case like this. And it is definetely possible.
If you choose to ignore this I think we should  have a jira task for it, otherwise it will will be forgotten in eternity ....eternity ... eternity _(weird echo)_

components/paragraphs/ParagraphResolver.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const type = paragraph?.__typename
+    if (!type) {
+      return null
+    }
+
+    const DynamicComponentType = components[type as keyof typeof components] || null
+    if (DynamicComponentType === null) return null
+
+    return (
+      <ParagraphErrorBoundary key={index}>
```
spaceo, line 23: Nice that you introduce a Paragraph Error boundary.
Is we talked about earlier we should have a clear defined way of treating errors throughout the application.
Please write an ADR about what the idea is behind treating errors in paragraphs. Or if you have a general idea of handling errrors in components with error boundaries the ADR can contain information about that too.

components/paragraphs/paragraphErrorBoundary/paragraphErrorBoundary.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { Button } from "../../shared/button/Button"
+
+function ErrorFallback({ resetErrorBoundary }: { resetErrorBoundary: () => void }) {
+  return (
+    <div className="bg-sand flex h-[200px] w-full flex-col items-center justify-center">
+      <p>Noget gik galt ved visning af denne blok - måske den mangler indhold?</p>
+      <Button className="mt-4" onClick={resetErrorBoundary}>
+        Genindlæs
+      </Button>
```
spaceo, line 13: Do we want a button for reloading failed content?

codegen.ts
spaceo: Actually it should not be needed when you have inserted this:
https://github.com/danskernesdigitalebibliotek/dpl-go/pull/94/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R12

So it should be either or.

package.json
```diff
@@ -5,19 +5,19 @@
...
-    "dev:https": "next dev --experimental-https",
-    "dev": "next dev",
+    "dev:https": "NODE_TLS_REJECT_UNAUTHORIZED=0 next dev --experimental-https",
+    "dev": "NODE_EXTRA_CA_CERTS=\"$(mkcert -CAROOT)/rootCA.pem\" next dev",
     "format:check": "prettier --check .",
     "format:write": "prettier --write .",
     "lint": "next lint",
     "start:with-server-source-maps": "NODE_OPTIONS='--enable-source-maps=true' next start",
-    "start": "next start",
+    "start": "NODE_EXTRA_CA_CERTS=\"$(mkcert -CAROOT)/rootCA.pem\" next start",
```
spaceo, line 20: This should not be inserted here, since we use `start` in. production

spaceo (changes requested):
I have some change requests and some areas to discuss

components/paragraphs/ParagraphGoVideo/ParagraphGoVideo.tsx
ThomasGross: Im pretty sure I implemented it because it had some buggy behavior on the demo site. But lets try to remove it and se if the bug is still present.

components/paragraphs/paragraphErrorBoundary/paragraphErrorBoundary.tsx
ThomasGross: I think if there's a onClick method on the element it should be a button. Maybe we could remove the button entirely? 

The resetErrorBoundary is in case you want to try reloading the element. Maybe a connection timeout or something happend and the paragraph did not load in time.

codegen.ts
ThomasGross: Im removing this line 😊

docs/architecture/adr-003-error-boundaries.md
```diff
@@ -0,0 +1,39 @@
...
+
+The implementation of error boundaries will be an ongoing process. We will integrate them progressively, adding them to components where they provide the most benefit as we continue to develop and enhance the site.
+
+## Alternatives considered
+
+none
+
+## Consequences
+
+none
```
spaceo, line 39: I think you could move the "Graceful Error Handling", "Component Isolation", "Debugging and Logging" etc into the "Consequences" section, because they describe the benefits about using error boundaries.
And the "Context" should describe what kind of problem field we are trying to solve.

spaceo (approved):
Greatnesss. LGTM
I think the ADR should be restructured. But otherwise things are fine and dandy

app/[...slug]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+async function page(props: { params: Promise<{ slug: string[] }> }) {
+  const params = await props.params
+
+  const { slug } = params
+
+  const data = await fetcher<GetPageByPathQuery, { path: string }>(GetPageByPathDocument, {
+    path: slug.join("/"),
+  })()
+
```
Comment (spaceo) [inline app/[...slug]/page.tsx:20]: Just spotted this. This is different than how we normally do it. And I think there is two problems here:
* We should use a common query client so we can control caching on a global scale if we want
* Consitency

We should do something like     return queryClient.fetchQuery({
      queryKey: useSearchWithPaginationQuery.getKey(args),
      queryFn: useSearchWithPaginationQuery.fetcher(args),
    })
.

spaceo (changes requested):
Just spotted something that needs to be changed

spaceo (approved):
LGTM 👍

PR #97, opened 2025-01-24
Bypass graphql cache in development

spaceo:
Bypass graphql cache in development
And set the staleTime to 1 minute in production because we, in dpl-cms content context, will rely on the caching system in NextJs


Adamik10 (approved):
MASTERFUL!

PR #106, opened 2025-01-29
Introducing a primary/secondary theme on buttons

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-355

#### Description

We want to be able to distinguish between buttons being more important than others depending on context by classifying as being primary or secondary.


ThomasGross (approved):
Beautiful! ✅

PR #108, opened 2025-01-30
Move dpl cms config out of go config

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-355

#### Description

This PR mainly removes dplcms / unilogin out of the goConfig system.

A bug was discovered that accidentally removed the code_verifier from the session coming back from Unilogin. It has been fixed as well.


ThomasGross (approved):
Looks good 😊 Great to see that external called variables have been separated out from the config and added through the queryclient setup

PR #109, opened 2025-01-30
Storybook env variables

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-423

#### Description

This PR contains functionality that ensures storybook stories has access to env variables. These variables needs to be added in to the workflow .yml file.


spaceo (approved):
Ship it!

PR #110, opened 2025-01-30
feat: restructure page loading logic to handle homepage and add URL resolver

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-428

#### Description

restructure page loading logic to handle homepage and add URL resolver


.env.test
```diff
@@ -1,3 +1,4 @@
+NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS=https://dapple-cms.docker/graphql
```
spaceo, line 1: Why is this needed?

lib/graphql/fetchers/dpl-cms.fetcher.ts
```diff
@@ -23,8 +23,12 @@ export function fetcher<TData, TVariables>(
...
 
-  if (!dplCmsGraphqlEndpoint || !dplCmsGraphqlBasicToken) {
-    throw new Error("Missing DPL CMS GraphQL endpoint or basic token")
+  // Check if the environment variables are set
+  if (!dplCmsGraphqlEndpoint) {
+    throw new Error("Missing DPL CMS GraphQL endpoint")
+  }
+  if (!dplCmsGraphqlBasicToken) {
+    throw new Error("Missing DPL CMS GraphQL basic token")
   }
```
spaceo, line 32: Thumbs up!

lib/config/resolvers/urls.ts
```diff
@@ -0,0 +1,9 @@
+const GO_FRONT_PAGE_SLUG = "go-frontpage"
+
+const urls = {
+  "url.gofrontpage": () => {
+    return GO_FRONT_PAGE_SLUG
+  },
+}
+
+export default urls
```
spaceo, line 9: We decided to call this scope: `route`

spaceo (approved):
We have decided upon some changes, when they are done it is good to go! 👍

.env.test
ThomasGross: Its added to because we want to align test env with local/example env where `NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS ` is present. 

The defined route is not necessary lets define it as "xxxx"

PR #111, opened 2025-01-31
refactor: Adjust mobile typography to match figma design

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-423

#### Description

This PR updates the mobile fonts to match the design


Adamik10 (approved):
I think some of the mobile fonts might be too small to be accessible. But our tests seem to be okay with it 🤷 so I guess it's good? 😅

PR #119, opened 2025-02-04
Improving dpl-cms data fetching

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-440

#### Description
The changes here was made in frustration of a weird behavior of fetching dpl-cms configuration in Go.
There will be an ADR describing the handling of cache in Go in general, but I think we need to handle the caching of the dpl-cms data via the NextJS caching system.
Later on we will look at implementing revalidation handling of the cache when changing content in the CMS, so it make sense to bypass the QueryClient cache of ReactQuery and solely rely on the Next cache system.

This PR:
* Hardens the handling of headers and next options in the dpl-cms fetcher
* Bumps up Next version
* Adds some debugging in the dpl-cms fetcher to be wiser of what is going on


ThomasGross (approved):
LGTM ✅

PR #127, opened 2025-02-18
Set NEXT_PUBLIC_APP_URL differently depending on environment

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-463

#### Description

We cannot use an extra "go." subdomain in pr environments so we need to use "go-" prefix instead.
This fix makes it possible to use the [nginx annotation setting ](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/3357a502ccf5bdc59a0c51f4f2b16b905c698259/.lagoon.yml#L171-L173) we need for setting te go-session cookie .

### DIFF
--- a/lagoon/start.sh
+++ b/lagoon/start.sh
@@ -10,7 +10,14 @@ getLagoonUrl() {
 }
 
 # Make sure app.url is set in application
-app_url=$(getLagoonUrl node)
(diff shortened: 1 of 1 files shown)


kasperbirch1 (approved):
LGTM 👍

PR #129, opened 2025-02-19
Ddfbra 337 jeg kan indsaette linkboks

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-442

#### Description

This include the linkbox component and stories


.env.example
```diff
@@ -1,4 +1,6 @@
 # YOUR DPL-CMS NEEDS TO BE RUNNING LOCALLY
+NEXT_PUBLIC_GRAPHQL_HOSTNAME=dapple-cms.docker
```
spaceo, line 2: I don't think that env variable is used anywhere?

.env.test
```diff
@@ -1,3 +1,4 @@
+NEXT_PUBLIC_GRAPHQL_HOSTNAME=xxxx
```
spaceo, line 1: I don't think that env variable is used anywhere?

app/api/getBase64/route.ts
```diff
@@ -0,0 +1,27 @@
+import { NextResponse } from "next/server"
```
spaceo, line 1: What is you idea of using the `api` in the route? 
How is it going to look like when we are introducing the Publizon API?

next.config.mjs
```diff
@@ -7,6 +10,11 @@ const nextConfig = {
         hostname: "res.cloudinary.com",
         pathname: "/**",
       },
+      {
+        protocol: "https",
+        hostname: env.NEXT_PUBLIC_GRAPHQL_HOSTNAME || "",
+        pathname: "/**",
```
spaceo, line 16: I don't understand the need for this?
We already have the `NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS`

next.config.mjs
```diff
@@ -24,4 +32,4 @@ const nextConfig = {
   },
 }
 
-export default nextConfig
+export default withPlaiceholder(nextConfig)
```
spaceo, line 35: typo

.eslintignore
```diff
@@ -1,2 +1,3 @@
 lib/rest/cover-service-api/generated/
 lib/rest/cover-service-api/generated/cover-service.ts
+next.config
```
spaceo, line 3: This has not been necessary to add so far. Why is it needed now?

spaceo (changes requested):
Great addition! 
But I have some questions I would like to have clarified.

next.config.mjs
ThomasGross: Jeg har tilføjet en pakke der har det navn 😊

.eslintignore
ThomasGross: Removed it. Seemed to be an issue earlier but now it seems ok.

PR #134, opened 2025-02-20
slider/videobundle updates and polish

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-481

#### Description

- Polish update to the material slider component
- Introduction of keen slider package
- implemented conditional dark/light check when using tailwind


components/shared/workCard/WorkCardWithCaption.tsx
```diff
@@ -1,30 +1,29 @@
...
   return (
     <div className={cn("block space-y-3 lg:space-y-5", className)}>
-      <WorkCard work={work} className={classNameWorkCard} isWithTilt={isWithTilt} />
-      <div className={cn("space-y-2", classNameCaption)}>
-        <p className="text-typo-subtitle-lg max-h-[72px] overflow-y-auto break-words">
-          {work.titles.full[0]}
-        </p>
-        <p className="text-typo-caption opacity-60">{displayCreators(work.creators, 2)}</p>
+      {children}
+
```
Adamik10, line 23:
```suggestion
```

components/shared/workCard/WorkCardStackedWithCaptionOld.tsx
```diff
@@ -0,0 +1,64 @@
+import Link from "next/link"
```
Adamik10, line 1: We should delete this file if it's old

Adamik10 (changes requested):
Very nice :)
I went through it and here are some general comments in addition to the file-specific ones:
- we should cut off the material slider content at the edge so it doesn't go all the way to the edge of the screen
- pls give the caption in work cards with caption in the material slider and in video bundle some minimal height so that the content doesn't jump around (my suggestion is so that it equals to 3 lines of the work title. That should be enough for most cases)
- when switching between light and dark mode, the background in work cards in material slider and in video bundle behaves weirdly. When going from light to dark - the transition is perfect, but the other way around it's hard-cutting between the colors.

ThomasGross:
> Very nice :) I went through it and here are some general comments in addition to the file-specific ones:
> 
> * we should cut off the material slider content at the edge so it doesn't go all the way to the edge of the screen
> * pls give the caption in work cards with caption in the material slider and in video bundle some minimal height so that the content doesn't jump around (my suggestion is so that it equals to 3 lines of the work title. That should be enough for most cases)
> * when switching between light and dark mode, the background in work cards in material slider and in video bundle behaves weirdly. When going from light to dark - the transition is perfect, but the other way around it's hard-cutting between the colors.

Im not a fan of the cut off:
[WorkCardWithCaption.tsx](https://github.com/danskernesdigitalebibliotek/dpl-go/pull/134/files#diff-d6c98c7b8f2ebf330c8140ca65584205192d6e12cd544a10ab29445c636ef93a)

The other points are fixed

Adamik10 (approved):
👍

PR #136, opened 2025-02-21
Make sure that cms url resolving looks after `cms-`

spaceo:
Since that is how we (at least for now) format the cms urls


kasperbirch1 (approved):
LGTM 👍

PR #137, opened 2025-02-21
Logout button and functionality (and documentation)

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-86

#### Description

This PR introduces a logout button/endpoint that works for both Unilogin and Adgansplaformen.
Furthermore the documentation of authentication in Go is added with the section about logging out as the first addition.

image


app/user/profile/LogoutButton.tsx
```diff
@@ -3,9 +3,16 @@
...
+  const session = useSession()
   const router = useRouter()
+
+  if (!session?.session?.isLoggedIn) {
+    return null
+  }
+
   return (
     <>
       <Button onClick={() => router.push("/auth/logout")} ariaLabel="Log ud">
```
Adamik10, line 18: In order to prevent similar problems as we had in the CMS regarding the logout button and session, try to move this callback function out of here into something like `const handleOnClick = () => {...}` , add a state to the component that tracks whether it's already been clicked or not. Then we can tie that state to `isDisabled` on the `<Button />` in order to prevent multiple clicks in short succession on that button.

docs/authentication.md
```diff
@@ -0,0 +1,47 @@
+# Authentication
```
Adamik10, line 1: Very nice with some documentation 👏

app/article/[id]/article.dpl-cms.graphql
Comment (Adamik10) [inline app/article/[id]/article.dpl-cms.graphql:1]: Maybe we shouldn't delete this folder? We decided to keep it when we talked earlier.

app/auth/logout/helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+  const appUrl = new URL(String(goConfig("app.url")))
+
+  return NextResponse.redirect(appUrl)
+}
+
+export const handleUniloginLogout = async (session: IronSession<TSessionData>) => {
+  const config = await getUniloginClientConfig()
+  const id_token = await getUniloginIdToken()
+
+  // If we don't have a client config, we can't do anything.
```
Adamik10, line 33: I really like these inline comments explaining this long procedure. Well done.

Adamik10 (commented):
Good job :) I have a couple of comments

app/article/[id]/article.dpl-cms.graphql
Comment (spaceo) [reply app/article/[id]/article.dpl-cms.graphql:1]: It was an old experiment. I think it is better to start on a clean slate. 👍

app/auth/logout/helpers.ts
spaceo: thx :)

PR #138, opened 2025-02-24
Put `LogoutButton` in suspense boundary

spaceo:
Since `useSession` (which is inside of the `LogoutButton`) uses `useSearchParams` we need to put the LogoutButton in a `Suspense` boundary.


Adamik10 (approved):
LGTM

PR #142, opened 2025-02-25
Move getBase64 route functionality to simple function

spaceo:
Unless I have misunderstood something there is no need for a round trip to a sepearte route for generating base64 hash of image.


ThomasGross (approved):
Looks good. Fix CI and merge 🙌

PR #144, opened 2025-02-26
Add article

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-326

#### Description
This PR adds article page, and related queries.

#### Screenshot of the result
-

-

### DIFF
--- a/app/[[...slug]]/page.tsx
+++ b/app/[[...slug]]/page.tsx
@@ -1,11 +1,10 @@
 import { notFound } from "next/navigation"
 import React, { Suspense } from "react"
 
+import loadPage from "@/app/[[...slug]]/loadPage"
 import BasicPageLayout from "@/components/pages/basicPageLayout/BasicPageLayout"
 import goConfig from "@/lib/config/goConfig"
 import { GetPageByPathQuery, NodeGoPage } from "@/lib/graphql/generated/dpl-cms/graphql"
-
-import loadPage from "./loadPage"
 
 async function page(props: { params: Promise<{ slug: string[] }> }) {
   const params = await props.params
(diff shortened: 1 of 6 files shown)


components/pages/articlePageLayout/ArticlePageLayout.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const mediaImage = goArticleImage?.mediaImage
+
+  const localeDateString = new Date(
+    pageData.publicationDate.timestamp as string
+  ).toLocaleDateString("da-DK", {
+    day: "numeric",
+    month: "long",
+    year: "numeric",
+  })
```
spaceo, line 27: I think we should:
* Make sure that the date is a Date object in the `loadArticle` function
* Introduce a date formatting helper in `/lib` with date formats we can use throughout the application
*  Use a formatting function from the helper here

spaceo (changes requested):
Looks good.
But I think we can handle the date formatting better

spaceo:
Btw. Just while I remember it:
Please include screenshots in a PR like this. It makes it easier to understand what is being implemented ;)

components/pages/articlePageLayout/ArticlePageLayout.tsx
```diff
@@ -18,13 +20,8 @@ function ArticlePageLayout({ pageData }: { pageData: TArticlePageLayoutProps })
...
 
-  const localeDateString = new Date(
-    pageData.publicationDate.timestamp as string
-  ).toLocaleDateString("da-DK", {
-    day: "numeric",
-    month: "long",
-    year: "numeric",
-  })
+  const publicationDate = new Date(pageData.publicationDate.timestamp)
+  const localeDateString = localeDateStringFromDate(publicationDate)
```
spaceo, line 25: I think we should choose a name that reflects the usage of the date. Since this is a date we will make use of throughout he application we could also call it `generalGoDate` or similar.

spaceo (approved):
Approved 👍 
I have one naming suggestion

PR #145, opened 2025-02-26
Rename insitution ids property

spaceo:
AMAZINGLY STIL all of a sudden renamed the property and Unilogin login was failing. This fixes the problem


Adamik10 (approved):
APPROVED

PR #148, opened 2025-02-28
Add local Publizon adapter - step 1

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-488

#### Description
This PR adds:
- local publizon adapter generated code
- local publizon adapter fetcher
- add the 1st publizon custom hook that determines whether to use local/external pubhub hook - useGetV1UserLoans  

#### Screenshot of the result
-

-


lib/rest/publizon/local-adapter/mutator/fetcher.ts
```diff
@@ -21,7 +21,7 @@ export const fetcher = async <ResponseType>({
 
   const body = data ? JSON.stringify(data) : null
   const serviceUrl = getRestServiceUrlWithParams({
-    baseUrl: "https://pubhub-openplatform.test.dbc.dk",
+    baseUrl: `${process.env.NEXT_PUBLIC_APP_URL}/pubhub`,
```
spaceo, line 24: I think we should define the route in our route config in `lib/config/resolvers/routes.ts`

spaceo (approved):
LGTM but please look at comment

PR #150, opened 2025-03-01
Makes sure that session is destroyed when expired

spaceo:
#### Link to issue

#### Description

Makes sure that session is destroyed when expired
Also tidied up strurture of session and logout functions.


Adamik10 (approved):
LGTM ✨

PR #151, opened 2025-03-01
Pubhub endpoint with user loans and session uniid

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-489

#### Description

Pubhub endpoint that uses the cardnumber from the session uniid.
The endpoint is prepared with one/all functionality that can load one or all entities of a resource.


Adamik10 (approved):
Nothing stands out to me as needing comments. Approved

PR #161, opened 2025-03-05
Unilogin user can view their loans

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-501

#### Description
This PR:
-  adds the possibility to view unilogin user's loans on the profile page - V1: we only show ISBNs for those works
- fixes up smaller UI issues around the app

#### Screenshot of the result
-

-


app/user/profile/page.tsx
```diff
@@ -1,16 +1,21 @@
...
 
 import DebuggingSession from "./DebuggingSession"
+import LoanList from "./LoanList"
 import LogoutButton from "./LogoutButton"
 
 const Page = () => {
   return (
-    <div className="content-container">
-      <h1 className="text-typo-heading-3 pb-5 font-bold">Profile</h1>
-      <Suspense fallback={<p>Loading...</p>}>
```
spaceo, line 10: I think the `< DebuggingSession> ` component requires the `<Suspense>` to be there. Unfortunately. But we just remove the suspense when we do not need the debugging anymore. It has to do with server/client relations and awaiting stuff and yada yada.
Maybe, if you would be so kind, could write a `@todo` about that we should remove the suspense when removing the debugging?

spaceo (approved):
LGTM 👍 
Just please reintroduce the suspense otherwise I am pretty sure that the user page will break

PR #163, opened 2025-03-05
Authorization check for Unilogin users

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-383

#### Description

This PR:
* Adds wsdl generation for Unilogin Soap services (first off we need the hentInstitution service)
* Adds a check for Unilogin users if they belong to an institution with an allowed municipality id
* Adds two errorpages: One for failed login and one for "not authorized login" for Unilogin users without access


.env.example
```diff
@@ -20,3 +20,5 @@ NEXT_PUBLIC_APP_URL=https://dapple-cms.docker:3000
 UNLILOGIN_PUBHUB_CLIENT_ID=XXX
 UNLILOGIN_PUBHUB_RETAILER_ID=XXX
 UNLILOGIN_PUBHUB_RETAILER_KEY_CODE=XXX
+UNLILOGIN_SERVICES_WS_USER=XXX
```
Adamik10, line 23: Could we pls add all these relevant things into our onePass? Maybe with the same naming so that we all know where to find them even if we can't reach you.

app/auth/callback/adgangsplatformen/route.ts
```diff
@@ -16,7 +16,8 @@ export async function GET() {
 
   // We could not retrieve the user token.
   // So we redirect to the frontpage without setting the session.
```
Adamik10, line 18: We don't redirect to the front page here right? Let's update this comment if that's the case.

components/pages/errorPageLayout/ErrorPageLayout.tsx
```diff
@@ -0,0 +1,18 @@
...
+const ErrorPageLayout = ({ lines }: { lines: (string | ReactNode)[] }) => {
+  return (
+    <div className="content-container flex items-center justify-center">
+      <div className="bg-white p-8 text-center shadow-md">
+        <h1 className="text-typo-heading-3 mb-6 font-bold">{lines[0]}</h1>
+        {lines.slice(1).map((line, index) => (
+          <p key={index} className="mb-6">
+            {line}
+          </p>
+        ))}
```
ThomasGross, line 12: I think this could be simpler and readable with using children prop and h1 as a title prop. 

```suggestion
        {children}
```

app/auth/callback/unilogin/route.ts
```diff
@@ -83,29 +101,14 @@ export async function GET(request: NextRequest) {
     // await session.save()
```
Adamik10, line 101: Is this commented out code necessary here?

lib/helpers/publizon.ts
```diff
@@ -0,0 +1,9 @@
+import { md5 } from "js-md5"
+
+export const getPublizonServiceParameters = () => {
+  return {
+    clientid: process.env.UNLILOGIN_PUBHUB_CLIENT_ID ?? "",
+    retailerid: process.env.UNLILOGIN_PUBHUB_RETAILER_ID ?? "",
+    retailerkeycode: md5(process.env.UNLILOGIN_PUBHUB_RETAILER_KEY_CODE ?? ""),
```
ThomasGross, line 7: Consider moving the hash function out where it is used.

Consider a comment on why we are using md5. 

Should these variables be added to the config instead of the helper folder? 
They don't necessarily need to be returned from a function as an object as they are constants and could be loaded individually where its needed.

lib/helpers/publizon.ts
```diff
@@ -0,0 +1,9 @@
+import { md5 } from "js-md5"
```
Adamik10, line 1: Why are we moving this away from the pubhub folder? I think it's nice that it was in the local helper.ts file.

Adamik10 (approved):
Nice job :) I have a couple question for you to look at before you merge this pls 🙏

.env.example
spaceo: Good point!

ThomasGross (changes requested):
I have a few comment, but very nice to see the progression. GJ 🙌

app/auth/callback/adgangsplatformen/route.ts
spaceo: 👍

.env.example
spaceo: Made a "Go env.local" entry

app/auth/callback/unilogin/route.ts
spaceo: That change I made too fast. Thx for catching that!

lib/helpers/publizon.ts
spaceo: Good questions.
I in general like a declarative way of coding. In this case I want a set of properties that I am pretty sure is a common set of parameters we will need for all of the Publizon SOAP request. So that is the purpose of the `getPublizonServiceParameters` function (could also be called: `getCommonPublizonServiceParameters`) I just want the parameters as a set because they are needed together repeatedly. If I want a single one of them I can just destruct that value. I also don't want to think about if one of the parameters needs to be transformed - in this case the retailerkeycode is md5 hashed.
I agree it could make sense to comment why the value is hashed, but apart from that I don't think that should be done more about the function.
The config question is also an interesting one.
The env variables has kind of been moving away from config. And frankly I am in doubt wether config should be used solely for static values (like search limit, routes etc) and env variables are treated separately. But maybe it would be nice to consolidate all the env variables in config. Maybe that is nice. But even so I would still need a declarative function that tells me the common parameters for the SOAP request.

components/pages/errorPageLayout/ErrorPageLayout.tsx
spaceo: We agreed to possibly create a future PR if we want the component structure to be changed

ThomasGross (approved):

PR #165, opened 2025-03-07
Fix inaccurate automatic accessibility test

JacobArrow:
#### Link to issue

[DDFBRA-448](https://reload.atlassian.net/browse/DDFBRA-448)

#### Description

When `duration-dark-mode` and `transition-all` is applied to the body element, it will also affect components in storybook, which causes the background to animate when loading a story. This causes axe to register the wrong colors, as colors are being animated while axe runs its checks.

This PR moved those mentioned classes to the Nextjs layout so stories will not have the transition effect.

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -42,7 +42,7 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="da">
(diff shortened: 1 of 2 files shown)


Adamik10 (approved):
Omg so good! Can't wait to have accessibility online again ❤️

PR #175, opened 2025-03-12
Unilogin user can loan a material

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-376

#### Description
This PR makes it possible for Unilogin users to loan e-materials.
As a part of that it introduces local Publizon API for app/pubhub/v1/user/loans/[...identifier]

#### Screenshot of the result
-

-

### DIFF
--- a/lib/config/env.ts
+++ b/lib/config/env.ts
@@ -40,7 +40,7 @@ const EnvSchema = z.object({
 // Should only be fetched with getServerEnv().
 const EnvServerSchema = z.object({
   UNILOGIN_API_URL: z.string().refine(validateUrl),
-  GO_SESSION_SECRET: z.string().length(32),
+  GO_SESSION_SECRET: z.string().min(32),
   UNLILOGIN_PUBHUB_CLIENT_ID: z.string(),
   UNLILOGIN_PUBHUB_RETAILER_ID: z.string(),
   UNLILOGIN_PUBHUB_RETAILER_KEY_CODE: z.string(),


__tests__/pubhub.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    })
+  })
+
+  it("Returns authorized for logged in users at GET /v1/user/loans", async () => {
+    vi.spyOn(sessionFunctions, "getSession").mockResolvedValue(
+      // @ts-ignore
+      Promise.resolve({
+        isLoggedIn: true,
+        type: "unilogin",
+        userInfo: { uniid: "100006cbab", institution_ids: ["A04441"] },
```
spaceo, line 37: Just to be sure: This uniid is not a real uniid from production right?

app/auth/callback/unilogin/route.ts
```diff
@@ -91,12 +93,15 @@ export async function GET(request: NextRequest) {
...
 
     // Set user info.
+    // TODO: After Publizon allows DDF test users to loan, we can remove thie if statement.
     session.userInfo = {
       sub: userinfo.sub,
       uniid: introspect.uniid,
-      institution_ids: introspect.institution_ids,
+      // TODO: Rename this into institutionIds
+      institution_ids:
+        institutionId === "A04441" ? ["101047"] : getInstitutionIds(introspect.institution_ids),
```
spaceo, line 102: Please write why we are doing this 😅

app/pubhub/helper.ts
```diff
@@ -1,3 +1,9 @@
 export const transformTimeToUtcString = (timeString: string) => {
   return timeString.split(" ").join("T")
 }
+
+export const regexDate = /^[0-9]{4}-[0-9]{2}-[0-9]{2}[ T][0-9]{2}:[0-9]{2}:[0-9]{2}[\.0-9A-Z]*$/
```
spaceo, line 5: A space sneaked in in front of the `T` in the middle. I don't know if it is a mistake?

app/pubhub/v1/user/loans/[identifier]/route.ts
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      expirationDateUtc: transformTimeToUtcString(response.data.expirationdateutc),
+      orderNumber: response.data.internalordernumber,
+      orderId: response.data.retailerordernumber,
+      code: response.status.code,
+      message: `${response.status.message} (#${response.status.code})`,
+    }
+  })
+  try {
+    const responseData = await createLoanRequest(context.userInfo, getLastValueFromUrl(request.url))
```
Comment (spaceo) [inline app/pubhub/v1/user/loans/[identifier]/route.ts:22]: You should get the identifier via the an argument in  the `postLibraryLoan` function. You can find documentation about it in the Nextjx documentation somewhere around dynamic routes.

app/pubhub/v1/user/loans/[identifier]/route.ts
```diff
@@ -0,0 +1,34 @@
+import { NextRequest, NextResponse } from "next/server"
+
+import { withAuth } from "@/app/pubhub/(lib)/helper"
+import { TUserInfo } from "@/app/pubhub/(lib)/types"
+import { getLastValueFromUrl, transformTimeToUtcString } from "@/app/pubhub/helper"
+
+import { createLoanRequest } from "./(lib)/requests"
+import { createLoanSchema } from "./(lib)/schemas"
+
+async function postLibraryLoan(request: NextRequest, context: { userInfo: TUserInfo }) {
```
Comment (spaceo) [inline app/pubhub/v1/user/loans/[identifier]/route.ts:10]: I think we should call the function `createLoan` or `createdigitalLoan`. The `POST` export later on shows that we are dealing with a POST request so we can instead focusing on calling our function something human readable :)

spaceo (commented):
Looks good 👏 
I have some comments to look at.
It seems like the PR is already pretty big. Is it possible to split up? Are all the changes dealing with: "Unilogin user can loan a material"?
If it is to big of a hazzle doing so then nevermind...

__tests__/pubhub.test.ts
Adamik10: It's the uniid for the test user with username "go" 😁

app/auth/callback/unilogin/route.ts
Adamik10: It is explained right above on line 96. Do you think we need more?

app/pubhub/helper.ts
Adamik10: It's unfortunately not a mistake. I found out that sometimes we can also get a date in that format - no T but a space instead. 🙄

app/pubhub/v1/user/loans/[identifier]/route.ts
Comment (Adamik10) [reply app/pubhub/v1/user/loans/[identifier]/route.ts:22]: Problem there ([see example](https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#example)) is that the useRouter is a hook and hooks can't be called inside functions, only components. Do you still think this is insufficient in that case?

app/auth/callback/unilogin/route.ts
spaceo: Sorry you're tight. Move along, nothing to see here.

__tests__/pubhub.test.ts
```diff
@@ -0,0 +1,87 @@
+// TODO: Test that the local adapter's output is the same as the adapter's output
+// We need two users with the same one loan
+// Mock soap service output
+// Copy output from the adapter and use it as assertion that our local adapter returns the same data
```
spaceo, line 4: Outdated comment right? A test was made and everything is fine, no?

app/pubhub/v1/user/loans/[identifier]/route.ts
```diff
@@ -0,0 +1,34 @@
...
+    const responseData = await createLoanRequest(context.userInfo, getLastValueFromUrl(request.url))
+    return NextResponse.json(createLoan.parse(responseData))
+  } catch (error) {
+    console.error(error)
+    throw new Response("Unprocessable content", { status: 422 })
+  }
+}
+
+// @todo Fix ts problem about context here.
+// @ts-ignore
```
Comment (spaceo) [inline app/pubhub/v1/user/loans/[identifier]/route.ts:31]: As far as I remember we have this comment in several places now. :) I think we should create a Jira task so we can get it solved.

components/pages/workPageLayout/LoanMaterialModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const [isHandlingLoan, setIsHandlingLoan] = useState(false)
+  const [error, setError] = useState<Error | null>(null)
+  const handleLoanMaterial = () => {
+    setIsHandlingLoan(true)
+    mutate(
+      { identifier: isbns[0] },
+      {
+        onSuccess: () => {
+          // Refetch data to update the UI for WorkPageButtons
+          queryClient.invalidateQueries({ queryKey: ["/v1/user/loans"] })
```
spaceo, line 64: Good that you remembered this!

components/pages/workPageLayout/LoanMaterialModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+          setIsHandlingLoan(false)
+          setIsOpen(false)
+        },
+        onError: err => {
+          setError(err as Error)
```
spaceo, line 69: I guess this would be handled differently when we get a global modal system

components/pages/workPageLayout/LoanMaterialModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+          {isLoadingLoans && (
+            <Icon
+              name="go-spinner"
+              ariaLabel="Indlæser"
+              className="animate-spin-reverse h-[24px] w-[24px]"
+            />
+          )}
+        </Button>
+      </div>
+    </ResponsiveDialog>
```
spaceo, line 160: I think the component is getting too big and it would be worth splitting into smaller components. 
Maybe it is not worth doing before it is transformed into using the global modal system, but then we could write a todo about it. 
I think it is getting difficult to get an overview of what is happening when a component is getting this big. Especially also because there is som much logic checking. 
By splitting the different cases into sub components makes it easier to understand the logic because it stands out. And then you can dive into the sub components if you want to have a closer look into the rendering of the fragments.

components/pages/workPageLayout/WorkPageButtons.tsx
```diff
@@ -68,45 +69,57 @@ const WorkPageButtons = ({ workId, selectedManifestation }: WorkPageButtonsProps
...
-            disabled={!!!identifier}
-            className="mb-grid-gap-half w-full lg:max-w-80 lg:min-w-72"
-            onClick={() => setIsPlayerOpen(!isPlayerOpen)}>
-            Prøv lydbog
-          </Button>
+          {!isLoaned && (
+            <Button
+              ariaLabel="Prøv lydbog"
+              size={"lg"}
+              disabled={!!!identifier}
```
spaceo, line 87: It is a matter of personal preference but I think `!!!!` is unreadable. To me `!Boolean(identifier)` is more clear.

components/pages/workPageLayout/WorkPageButtons.tsx
```diff
@@ -119,11 +132,17 @@ const WorkPageButtons = ({ workId, selectedManifestation }: WorkPageButtonsProps
...
             setIsPlayerOpen(!isPlayerOpen)
           }}
-          title={`Prøv ${isManifestationPodcast(selectedManifestation) ? "Podcast" : "Lydbog"}`}
-          description={`For at låne ${isManifestationPodcast(selectedManifestation) ? "podcast" : "lydbogen"} skal du være oprettet som bruger på GO.`}>
+          title={`Prøv ${isManifestationPodcast(selectedManifestation) ? "Podcast" : "Lydbog"}`}>
           <Player type="demo" identifier={selectedManifestation.identifiers[0].value} />
         </ResponsiveDialog>
       )}
+
+      {/* "Are you sure?" dialog */}
```
spaceo, line 140: Nice with the comment 👍

components/pages/workPageLayout/helper.ts
```diff
@@ -18,8 +21,23 @@ export const getWorkMaterialTypes = (
...
+    return "e-bog"
+  }
+  if (isManifestationAudioBook(manifestation)) {
+    return "lydbog"
+  }
+  if (isManifestationPodcast(manifestation)) {
+    return "podcast"
+  }
+  return undefined
 }
```
spaceo, line 41: I prefer to return `null` in a case like that. By returning `null` you explicitly tell that you have made the choice of doing so. 
So it is like saying:
We don't know how to resolve the type so we return `null` because it is "our" value and not because something slips through the system from an API call. In that way it is possible to distinguish between some thing that has been set deliberately or an unknown value that has not been "catched".
Does it make sense?

components/pages/workPageLayout/helper.ts
```diff
@@ -101,3 +119,55 @@ export const getIconNameFromMaterialType = (materialType: GeneralMaterialTypeCod
...
+    }
+  }) as SlideSelectOption[]
+}
+
+export const sortSlideSelectOptions = (options: SlideSelectOption[]) => {
+  return options.sort((a, b) => {
+    // sort by the index of the GeneralMaterialTypeCodeEnum in the materialTypeSortPriority array
+    return (
+      goConfig("materialtypes.sortpriority").indexOf(a.code) -
+      goConfig("materialtypes.sortpriority").indexOf(b.code)
```
spaceo, line 139: You could choose to extract the goConfig once before the sort function. But it is not a biggie and there should be no performance issues bt doing what you have done here.

components/pages/workPageLayout/helper.ts
```diff
@@ -101,3 +119,55 @@ export const getIconNameFromMaterialType = (materialType: GeneralMaterialTypeCod
...
+    )
+  })
+}
+
+export const getManifestationMaterialTypeIcon = (manifestation: ManifestationWorkPageFragment) => {
+  const materialType = getManifestationMaterialType(manifestation)
+  return getIconNameFromMaterialType(materialType.code) || "book"
```
spaceo, line 146: Is it possible to use a constant, either from our code or from generated code instead of just the "random" string `"book"`?

components/shared/workCard/MaterialTypeIconWrapper.tsx
```diff
@@ -1,21 +1,25 @@
 import React from "react"
 
+import Icon from "@/components/shared/icon/Icon"
 import { cn } from "@/lib/helpers/helper.cn"
 
-import Icon from "../icon/Icon"
+export type MaterialTypeIconWrapperProps = {
+  iconName: "ebook" | "book" | "headphones" | "controller" | "video" | "podcast"
```
spaceo, line 7: I think we should move this to our global type definitions in the global lib folder

spaceo (changes requested):
A lot of changes here 😅 
Please look at my comments

app/pubhub/v1/user/loans/[identifier]/route.ts
Comment (Adamik10) [reply app/pubhub/v1/user/loans/[identifier]/route.ts]: I did that now :))

components/pages/workPageLayout/LoanMaterialModal.tsx
Adamik10: I am writing a todo and making a ticket

components/pages/workPageLayout/WorkPageButtons.tsx
Adamik10: I can change that

components/pages/workPageLayout/helper.ts
Adamik10: Yesss I changed it to null :))

components/pages/workPageLayout/helper.ts
Adamik10: I can write a comment on why we return "book" 😉

spaceo (approved):
LGTM 👍

PR #176, opened 2025-03-12
Added new way of handlings env variables

JacobArrow:
#### Link to issue

[DDFBRA-370](https://reload.atlassian.net/browse/DDFBRA-370)

#### Description

This PR adds a unified way of handling env variables.
Many env variables are required to run the application properly, and sometimes arbitrary errors will accour if specific variables doesnt exist or follow a required type: eg. when iron-session requires a password to be at least 32 characters long.

By having a single place to store and access our application wide env variables, we can quickly identify if our .env.local file does not live up to the expected schema required by the application. This also helps with IDE autocompletion, rename issues and accidental removed variables.

- Go through each commit when reviewing

### DIFF
--- a/docs/architecture/adr-004-env-variables.md
+++ b/docs/architecture/adr-004-env-variables.md
@@ -26,7 +26,8 @@ Example:
 ```typescript
 function getEnvs() {
   return {
-    NEW_ENV: process.env.NEW_ENV,
+    NEW_ENV: process.env.NEXT_PUBLIC_NEW_ENV,
+    SOME_API_KEY: process.env.SOME_API_KEY,
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


lib/config/dpl-cms/dplCmsConfig.ts
```diff
@@ -9,18 +11,27 @@ const queryDplCmsConfig = async () => {
...
+
+  // If all env vars are present, return them directly without fetching config
+  if (wellknownUrl && clientId && clientSecret) {
+    return {
+      wellknownUrl,
+      clientId,
+      clientSecret,
+      apiData: null,
+    }
+  }
```
spaceo, line 26: I still think that we can do this in a partial way / merging values from env and fetch.
How about that we:
* Get the values as you do in line 14-16
*.Make an object out of it and eg. call it `configEnv`
* Fetch the API config with a try/catch block so we do not react on network/malformed errors
* Create an object with the API values eg. call it `configAPI`. If an error occured fetching the data the object will be empty
* Return `{...configAPI, ...configEnv}`

Could that make sense?

lib/helpers/unilogin.ts
```diff
@@ -13,9 +11,7 @@ export const getUniloginWsCredentials = () => {
...
 
 export const getLibraryMunicipalityId = () => {
-  const municipalityId = process.env.UNILOGIN_MUNICIPALITY_ID ?? ""
-  if (!municipalityId) {
-    throw new Error("Missing Unilogin WS credentials")
-  }
+  const municipalityId = getServerEnv("UNILOGIN_MUNICIPALITY_ID")
+
   return municipalityId
 }
```
spaceo, line 17: Just a sugesstion. More a matter of taste. But since there is not a lot of meat/logic in this function I would write:
```typescript
export const getLibraryMunicipalityId = () => getServerEnv("UNILOGIN_MUNICIPALITY_ID")

spaceo (changes requested):
Great addition :) 👍 
I just think we should fix the api/env patching problem I wrote about.

lib/config/dpl-cms/dplCmsConfig.ts
JacobArrow, line 1: Like this @spaceo ?

spaceo (approved):
Great. LGTM 👍

PR #186, opened 2025-03-25
Add ADR-005 Dual login and Publizon Adapter

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-380

#### Description
This PR adds an architectural decision record regarding dual login and publizon adapter.

#### Screenshot of the result
-

-


docs/architecture/adr-005-dual-login-and-publizon-adapter.md
```diff
@@ -0,0 +1,67 @@
...
+## Context
+
+Our application supports two different authentication mechanisms for users:
+
+- **Adgangsplatformen** — used by library patrons.
+- **Unilogin** — a login used by students, particularly children in schools.
+
+Depending on which method is used, the session is established differently and includes a session token that is used to authorize requests to external services.
+
+A key integration in our system is the **Publizon Adapter** ([see documentation here](https://library-api.qa.pubhub.dk/index.html)), a service that acts as a middle layer between us and the **Publizon Pubhub** (a SOAP-based backend-to-backend system). [See documentation here.](https://libraryservices.pubhub.dk/v2_7/) This adapter provides endpoints we use to achieve various things, such as:
```
spaceo, line 12: In our research we found out that the Publizon adapter actually is a very thin layer proxying request further to a Publizon API. So it is not correct that you write that it is a SOAP-based backend-to-backend system

docs/architecture/adr-005-dual-login-and-publizon-adapter.md
```diff
@@ -0,0 +1,67 @@
...
+
```
spaceo, line 13: I think it would be worth mentioning somwhere in this document that by reverse enginering the Ereolen Go functionality for Unilogin users we found out that SOAP services where used on the server side. And we generate clients for the various services that we use in our local Pubhub API.

spaceo (changes requested):
Great work 👏 
I have one correction and a suggestion about additional information.

docs/architecture/adr-005-dual-login-and-publizon-adapter.md
Adamik10: I think maybe we mean the same thing here. But if it's confusing to understand, I can reformulate it.

PR #187, opened 2025-03-26
feat: Add hardcoded childrenOrAdults filter to search queries

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-485

#### Description

This PR introduces a hardcoded filter to search results that ensures that only materials for children is fetched


spaceo (approved):
LGTM 👍

PR #188, opened 2025-03-26
Ddfbra 533 test om sti ls info service kan give adgang for brugerne

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-533

#### Description

- This PR handles has_licens which ensures the logged in user has the correct access
- Aligns all previous error pages to have same layout


spaceo (approved):
LGTM 👍

PR #191, opened 2025-03-31
Ddfbra 529 restructure modal system

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-529

#### Description

- Introduces a dynamic modal system
- Handling of opening of the different types of modals
- Fixes to loan logic and how the affected modal is shown because of that


__tests__/__snapshots__/pubhub.test.ts.snap
```diff
@@ -13,7 +13,7 @@ exports[`Pubhub local API > Returns same output from local & external GET /v1/us
       },
       "loanExpireDateUtc": "2025-04-10T14:28:57",
       "orderDateUtc": "2025-03-11T13:30:04",
-      "orderId": "7cab0a54-2a22-4172-862c-b3ed346e1da0",
+      "orderId": "5f9dacba-1ad5-4275-b90f-eac84a53ccae",
```
spaceo, line 16: This change is not really necessarry, right. I mean it does no harm, but it could have been left out.

components/pages/workPageLayout/WorkPageButtons.tsx
```diff
@@ -1,6 +1,5 @@
-import React, { useMemo, useState } from "react"
+import React, { useMemo } from "react"
```
spaceo, line 1: I think the`WorkPageButtons` component is too big and its content should be divided into several components. But we already talked about that, so there is no need to dwell into that.

components/pages/workPageLayout/WorkPageButtons.tsx
```diff
@@ -9,13 +8,14 @@ import {
...
 import { Button } from "@/components/shared/button/Button"
 import Icon from "@/components/shared/icon/Icon"
-import Player from "@/components/shared/publizonPlayer/PublizonPlayer"
-import ResponsiveDialog from "@/components/shared/responsiveDialog/ResponsiveDialog"
 import SmartLink from "@/components/shared/smartLink/SmartLink"
+import useSession from "@/hooks/useSession"
 import { ManifestationWorkPageFragment } from "@/lib/graphql/generated/fbi/graphql"
 import { cn } from "@/lib/helpers/helper.cn"
 import { resolveUrl } from "@/lib/helpers/helper.routes"
 import useGetV1UserLoans from "@/lib/rest/publizon/useGetV1UserLoans"
```
spaceo, line 16: We should only load `useGetV1UserLoans`if the user is not anonymous. I don't think that parts is handled right?

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -73,12 +76,20 @@ const LoanMaterialModal = ({ isOpen, setIsOpen, manifestation }: LoanMaterialMod
...
 
+  const { data: publizonData } = useGetV1ProductsIdentifierAdapter(isbns?.[0], {
+    query: {
+      // Publizon / useGetV1ProductsIdentifier is responsible for online
+      // materials. It requires an ISBN to do lookups.
+      enabled: isbns.length > 0,
+    },
+  })
+
+  const isLoanPossible =
```
spaceo, line 87: A litte comment here would be nice :)

components/shared/publizonReader/PublizonReader.tsx
```diff
@@ -38,26 +38,25 @@ const Reader = ({ type, onBackCallback, identifier, orderId }: ReaderType) => {
...
-          aria-label="Go back"
-        />
-      </div>
+      <div
+        style={{ height: "100vh" }}
+        id="pubhub-reader"
+        order-id={orderId}
+        role="button"
+        tabIndex={0}
+        // eslint-disable-next-line no-script-url
```
spaceo, line 49: Usually we write a comment when we bypass our own rules. I think we should do that here as well.

spaceo (changes requested):
Great additions!
I have some comments to be looked at though.

__tests__/__snapshots__/pubhub.test.ts.snap
ThomasGross: When running the test it failed if I did not change the orderId.

components/pages/workPageLayout/WorkPageButtons.tsx
ThomasGross: This should be separated out now

spaceo (approved):
LGTM 👍

PR #193, opened 2025-04-01
Ddfbra 378 indlogget unilogin bruger kan se sine lan the sequel

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-378

#### Description
This PR makes it possible for a unilogin user to see their loans and quotas.

#### Screenshot of the result
-

-


components/shared/coverPicture/CoverPictureWithCaption.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const CoverPictureWithCaption = ({
+  className,
+  children,
+  caption,
+}: CoverPictureWithCaptionProps) => {
+  return (
+    <div className={cn("block h-full w-full space-y-3 px-[15%]", className)}>
+      <div className="reltive h-[85%]">{children}</div>
```
ThomasGross, line 18: typo

```suggestion
      <div className="relative h-[85%]">{children}</div>
```

components/shared/coverPicture/CoverPictureWithCaption.tsx
ThomasGross, line 1: Consider if it's necessary to create a new component when its only used on the login page?

app/user/profile/ProfilePageLayout.tsx
```diff
@@ -0,0 +1,25 @@
+"use client"
```
ThomasGross, line 1: This could be server component right?

```suggestion
```

app/user/profile/ProfilePageLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div className="content-container grid-go -mt-space-y w-full space-y-3">
+      <div className="col-span-full flex flex-row flex-wrap">
+        <h1 className="text-typo-subtitle-sm mb-5 lg:w-full">Profile</h1>
+        <LogoutButton className="ml-auto justify-end lg:order-2" />
+        <p className="text-typo-heading-2 mt-6 w-full pb-5 lg:order-1 lg:mt-0 lg:w-auto lg:max-w-[80%]">
+          Username / User Name
+        </p>
+      </div>
+      <LoanList />
+      <DebuggingSession />
```
ThomasGross, line 20: This should be removed before going into main 

```suggestion
      <DebuggingSession />
```

app/user/profile/LoanList.tsx
ThomasGross, line 1: Idea: 
Call this component UserLoans / UserLoansSection / LoansSection

app/user/profile/LoanListSlider.tsx
ThomasGross, line 1: Idea:
Call this component LoanSlider

app/user/profile/LoansDetails.tsx
ThomasGross, line 1: This could be called quotaSection?

Consider moving this component to the ProfilePageLayout after loanlist

app/user/profile/LoansDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <div className="bg-background col-span-6 w-full space-y-6 rounded-sm px-10 pt-6 pb-9 md:w-[49%]">
+        <h3 className="text-typo-subtitle-sm col-span-full opacity-50">Kvote</h3>
+        <div className="align-center flex w-full flex-row justify-between">
+          <div className="bg-background-overlay flex h-36 w-[47%] flex-col items-center justify-center gap-4 rounded-sm">
+            <p className="text-typo-heading-3">
+              {loanData.userData?.totalEbookLoans} af{" "}
+              {loanData.libraryData?.maxConcurrentEbookLoansPerBorrower ??
+                Number(loanData.libraryData?.maxAmountPerMonth) -
+                  Number(loanData.libraryData?.maxConcurrentAudiobookLoansPerBorrower) ??
+                "?"}
```
ThomasGross, line 23: Consider moving this logic to a variable before the return

app/user/profile/LoansDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+            </p>
+            <p className="text-typo-subtitle-sm opacity-50">E-bøger</p>
+          </div>
+          <div className="bg-background-overlay flex h-36 w-[47%] flex-col items-center justify-center gap-4 rounded-sm">
+            <p className="text-typo-heading-3">
+              {loanData.userData?.totalAudioLoans} af{" "}
+              {loanData.libraryData?.maxConcurrentAudiobookLoansPerBorrower ??
+                Number(loanData.libraryData?.maxAmountPerMonth) -
+                  Number(loanData.libraryData?.maxConcurrentEbookLoansPerBorrower) ??
+                "?"}
```
ThomasGross, line 33: Consider moving this logic to a variable before the return

ThomasGross (changes requested):
Good work! 🙌 

I commented some ideas for adjustment and consideration. Please have a look. 😊

app/user/profile/LoanCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+      </div>
+    )
+  }
+
+  return (
+    <div className={cn("relative flex aspect-5/7 h-full w-full", className)}>
+      <div className="aspect-1/1 h-full w-full">
+        <CoverPictureWithCaption caption={`Udløber om ${daysUntil} dage`}>
+          <CoverPicture
```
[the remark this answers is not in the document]
Adamik10: I personally think it's cleaner and easier to read this way, where if anything is loading, we don't even get to the proper return(), but I can move it.

app/user/profile/LoanCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+            src={coverSrc?.[0] || ""}
+            alt={`${title} cover billede`}
+            withTilt={false}
+            className="select-none"
+          />
+          <MaterialTypeIconWrapper
+            iconName={getManifestationMaterialTypeIcon(
+              manifestation as ManifestationWorkPageFragment
+            )}
+            className="bg-background z-floating-icon relative mx-auto -mt-14 outline-1"
```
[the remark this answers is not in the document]
Adamik10: These are two different versions of it due to placement. The same version as the one on the work card wouldn't work due to the loan card not actually being a card, just a floating cover without a background behind it. 
But this one is the same as the one you see when you're loaning a material on the work page in the modal. I added an outline to that one too so that they have the same visual identity.

app/user/profile/LoanCard.tsx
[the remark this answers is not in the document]
Adamik10: don't you think it's cleaner to have it divided like I did it? I think it's easier to comprehend what's going on in the component. I think it gets a bit confusing with inlining the loading part into the main return()
But if you have some more technical arguments to the approach I am all ears :heart_hands:

components/shared/coverPicture/CoverPictureWithCaption.tsx
Adamik10: Does it have any downsides?

app/user/profile/ProfilePageLayout.tsx
Adamik10: It's been there since forever ago, and I think we should keep it until we implement the username. Me and Mikkel were talking yesterday and he said we should use the session for getting the username. So let's keep it until then.

app/user/profile/LoansDetails.tsx
Adamik10: It's visually a part of the slider, so I'd like to keep it located as is.

app/user/profile/LoansDetails.tsx
Adamik10: I actually spotted a flaw there, so now there's no need for those long lines 😉

app/user/profile/LoansDetails.tsx
Adamik10: I actually spotted a flaw there, so now there's no need for those long lines 😉

components/shared/coverPicture/CoverPictureWithCaption.tsx
ThomasGross: The component might be confusing, as its name suggests it is meant for general use, but it is actually tied to one very specific purpose on the loan card.

ThomasGross (approved):
Please look at the comments again. Consider if its something worth changing.

PR #195, opened 2025-04-03
Added vitest helper function to run tests silently

JacobArrow:
Sometimes we expect a test to throw errors, which sometimes also logs messages to the console. This will lead to confusion, because it looks like unitended errors while testing.
With this custom function we can silence those expected messages.

### DIFF
--- a/__tests__/env.test.ts
+++ b/__tests__/env.test.ts
@@ -2,6 +2,8 @@ import { beforeEach, expect, test } from "vitest"
 
 import { getEnv, getServerEnv } from "@/lib/config/env"
 
+import { testSilently } from "./helpers"
+
 beforeEach(() => {
   vi.unstubAllEnvs()
 })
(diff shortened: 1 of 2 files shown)


spaceo (commented):
I like this addition @JacobArrow! 
It is just important somehow (which is not a part of this PR) to make it clear for the dev team that `testSilently`is something you attached when you have made sure that everything works as expected.
But I must say that I have a more serious concern of mine:
Lets say that you have made test. You have silenced the test because you are not interested in showing the errors in the output. Now the codebase evolves and a new error occurs but you miss it because you have silenced the test - and that kind of sucks.
Maybe it is possible to add a count of expected errors so if additional errors occurs we get aware of it?

JacobArrow:
@spaceo `testSilently` does not prevent the application from throwing errors, nor does it prevent vitest from failing when an error occours. The only thing that it does, is that it prevents console.log and console.error from reaching the user when testing. 
Vitest does not care about console messages. They are only there to help a real person understand what is going on. 
If changes to the test or the codebase would introduce application errors, the vitest would fail as it should, but to better debug errors, one could replace `testSilently` with `test`.

In the end we only look at vitest console messages when something fails.

PR #196, opened 2025-04-03
Handle empty state for user loans

Adamik10:
#### Link to issue

#### Description
This PR handles the first part of the ticket only - empty state of the loan list. I am trying to divide the work so that I won't end up with a giant PR. Adding username to the page will be in its separate PR.

#### Screenshot of the result
-

-

### DIFF
--- a/app/user/profile/LoanSlider.tsx
+++ b/app/user/profile/LoanSlider.tsx
@@ -65,22 +65,24 @@ const LoanSlider = ({ works, loanData }: LoanSliderProps) => {
         <h2 className="text-typo-heading-4">
           E-materialer jeg har lånt ({loanData.loans?.length})
         </h2>
-        <div className="flex flex-row justify-end gap-x-4">
-          <Button
-            disabled={reachedStart}
-            variant="icon"
-            ariaLabel="Vis forrige værker"
-            onClick={() => onLeftClick()}>
-            <Icon className="h-[24px] w-[24px]" name="arrow-left" />
-          </Button>
-          <Button
-            disabled={reachedEnd}
-            variant="icon"
-            ariaLabel="Vis næste værker"
-            onClick={() => onRightClick()}>
-            <Icon className="h-[24px] w-[24px]" name="arrow-right" />
-          </Button>
-        </div>
+        {!!loanData.loans?.length && (
+          <div className="flex flex-row justify-end gap-x-4">
+            <Button
+              disabled={reachedStart}
+              variant="icon"
+              ariaLabel="Vis forrige værker"
+              onClick={() => onLeftClick()}>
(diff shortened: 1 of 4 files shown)


app/user/profile/LoanSlider.tsx
```diff
@@ -65,22 +66,24 @@ const LoanSlider = ({ works, loanData }: LoanSliderProps) => {
...
-          </Button>
-          <Button
-            disabled={reachedEnd}
-            variant="icon"
-            ariaLabel="Vis næste værker"
-            onClick={() => onRightClick()}>
-            <Icon className="h-[24px] w-[24px]" name="arrow-right" />
-          </Button>
-        </div>
+        {!!loanData.loans?.length && (
```
spaceo, line 69: You might as well just type:
```typescript
{loanData.loans?.length && (
Comment
```
because you just need to check the truthiness of the statement:
* if the `loans` property on `loanData.loans` is `undefined`it is falsy
* if the `loanData.loans.length`is 0 it is falsy

To me writing a double exclamation mark makes me wonder for a longer time what is actually going on.

app/user/profile/LoanSlider.tsx
```diff
@@ -108,6 +111,47 @@ const LoanSlider = ({ works, loanData }: LoanSliderProps) => {
               </Link>
             )
           })}
+          {works.length < 2 && (
```
spaceo, line 114: Would be nice with a comment about the works length in the varoius places of the rendered output.
Maybe before the return statement?

app/user/profile/LoanSlider.tsx
```diff
@@ -108,6 +111,47 @@ const LoanSlider = ({ works, loanData }: LoanSliderProps) => {
...
+                    className="min-w-80">
+                    {!isLoadingRedirect && "Find din næste bog"}
+                    {isLoadingRedirect && (
+                      <Icon
+                        name="go-spinner"
+                        ariaLabel="Indlæser"
+                        className="animate-spin-reverse mx-6 h-[15px] w-[15px]"
+                      />
+                    )}
+                  </Button>
```
spaceo, line 150: Could you maybe explain what `loadingRedirect` is all about?

spaceo (commented):
I have some comments.
Nothing big just clarifications

app/user/profile/LoanSlider.tsx
Adamik10: I had to reverse me changing this, because when the length of the array is 0 as a number, it renders the 0 in the frontend. So we do need the !! (boom! boom! you're a boolean) operator
image

spaceo (approved):
LGTM 👍

PR #197, opened 2025-04-03
Add server side mocking to Cypress

JacobArrow:
#### Link to issue

[DDFBRA-524](https://reload.atlassian.net/browse/DDFBRA-524)

#### Description

We need to have a consistent way of mocking server side data in Nextjs. This PR adds a custom [MSW](https://mswjs.io/) node mocking server, which startup together with Cypress before tests are run. From inside a Cypress test, it is now possible to order MSW on how to respond to request to specific endpoints or GraphQL operation names. MSW will listen to all application wide requests, and intercept them, if they match the specified rules and return whatever data provided. 

This is still an early step for the overall handling of server side requests. 
I've chosen to only focus on solving the mocking of Adgangsplatformen Token, to have this PR focus on the implementation and usage of MSW.


ThomasGross (approved):
LGTM ✨👍

PR #200, opened 2025-04-04
Common way of handling "loading" buttons

spaceo:
#### Link to issue

#### Description

Introducing a common way of handling loading state when clicking buttons that triggers longer loading processes.
And implements it on:
* Logout button
* "Find a book" button
* Login button

### DIFF
--- a/app/user/profile/LogoutButton.tsx
+++ b/app/user/profile/LogoutButton.tsx
@@ -1,9 +1,9 @@
 "use client"
 
 import { useRouter } from "next/navigation"
-import { useState } from "react"
 
 import { Button } from "@/components/shared/button/Button"
+import ButtonWithLoadingStateHoc from "@/components/shared/button/ButtonWithLoadingStateHoc"
 import Icon from "@/components/shared/icon/Icon"
 import useSession from "@/hooks/useSession"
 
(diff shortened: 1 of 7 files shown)


components/shared/button/ButtonWithLoadingStateHoc.tsx
```diff
@@ -0,0 +1,60 @@
...
+        </Button>
+      )
+    }
+    return (
+      <Component
+        {...(props as TProps & {
+          isLoading: boolean
+          onClick: typeof handleClick
+        })}
+        isLoading={isLoading}
```
Adamik10, line 53: We talked about removig this from the non-loading component + line 21 - const [buttonWasClicked, setbuttonWasClicked] = useState(false)

app/user/profile/LogoutButton.tsx
```diff
@@ -14,16 +14,17 @@ export const LogoutButtonSkeleton = () => (
...
+  onClick?: () => void
 }
+const className = "ml-auto lg:order-2 min-w-40"
+const size = "sm"
 
-const LogoutButton = ({ className }: LogoutButtonProps) => {
-  const [logoutWasClicked, setLogoutWasClicked] = useState(false)
-  const { session, isLoading } = useSession()
+const LogoutButton = ({ onClick }: LogoutButtonProps) => {
+  const { session, isLoading: sessionIsLoading } = useSession()
```
Adamik10, line 24: We should handle this on the user profile page, so we never actually see this button and have to make this check here.

Adamik10 (approved):
Approved with some comments, pls look :)))

PR #201, opened 2025-04-07
Finialize token/session management in middleware

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-462

#### Description

This PR makes sure that all session refreshing and expiration is handled in the middleware.
And following test are made:

     ✓ can ensure that a library token is present if it is not already
     ✓ can destroy an Adgangsplatformen session if it is active and a Drupal session does not exist
     ✓ can create an Adgangsplatformen session if it is not already created and a dpl-cms ession exists
     ✓ can refresh an Adgangsplatform session if it is expired
     ✓ does NOT refresh an Adgangsplatform session if it isn't expired yet
     ✓ can refresh a Unilogin session if it is expired
     ✓ can destroy a Unilogin session if the refresh time is overdue
     ✓ can destroy an Adgangsplatformen session if the access token lifetime has run out

#### How to review
I suggest taking a look at the [middleware function 
](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/298370deb5ab0004aa26a0d0a1d80b3c57037103/middleware.ts#L20)

And look at the test, starting with the test titles since they explain what is being handled.


lib/helpers/library-token.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import { getEnv } from "../config/env"
+import goConfig from "../config/goConfig"
+import getQueryClient from "../getQueryClient"
+import {
+  GetAdgangsplatformenLibraryTokenQuery,
+  useGetAdgangsplatformenLibraryTokenQuery,
+} from "../graphql/generated/dpl-cms/graphql"
+
+// export const libraryTokenExist = async () => {
```
Adamik10, line 14: Should this be deleted?

Adamik10 (approved):
Ufff this is so big. I have two comments, but other than that I couldn't find anything to improve this.

lib/helpers/library-token.ts
spaceo: Yes. Well spotted 👍

__tests__/middleware.test.ts
```diff
@@ -0,0 +1,386 @@
...
+
+    try {
+      await middleware(new NextRequest("http://localhost"))
+    } catch (e) {
+      expect(e).toEqual("Refresh token expired")
+    }
+    expect(refreshTokenGrantSpy).toHaveBeenCalledTimes(1)
+    expect(destroySessionSpy).toHaveBeenCalledTimes(1)
+  })
+
```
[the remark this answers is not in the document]
spaceo: Yes. Done.

PR #205, opened 2025-04-08
Pass original request headers through to service

spaceo:
Then we do not need to have any content-type header eg. logic. We simply pass along what we get.

### DIFF
--- a/app/ap-service/[[...slug]]/route.ts
+++ b/app/ap-service/[[...slug]]/route.ts
@@ -27,6 +27,11 @@ async function proxyRequest(
   { params }: TContext,
   body?: string
 ) {
+  const proxiedHeaders: Record<string, string> = {}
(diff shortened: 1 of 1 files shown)


Adamik10 (commented):
What problem does this PR solve?

spaceo:
> What problem does this PR solve?

It solves what I write in the description of this PR. But maybe I can be more clear :) :
Some of our fetchers uses the ap-service endpoint. The fetchers knows what kind of headers that needs to be sent to the various services. For instance the fbi fetcher knows that the content type header is "application/json". So instead of "guessing" the headers in the ap-service endpoint that performs the actual request, we use the headers that are already being sent by the fetcher to the ap-service. Does it make sense?

Adamik10 (approved):
Approved! Thx for explanation 😉

PR #208, opened 2025-04-10
Make sure that user name information is present in the go session

spaceo:
#### Link to issue

#### Description

Makes sure that user name information is present in the go session
Needed to refactor `userInfo` to `uniloginUserInfo `as well to emphasize that it is user info tied to a unilogin user.
Also the username (either name or username) is rendered instead of the temporary placeholder.


ThomasGross (approved):
A minor thing to consider otherwise perfect 👌

PR #211, opened 2025-04-14
Fix/handle unwanted manifestation types

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-485

#### Description

- filter out unwanted manifestations on each material type
- cleanup code and create helper functions for reused code


.env.production
```diff
@@ -8,6 +8,8 @@ NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_FBI=https://fbi-api.dbc.dk/ereolgo/graphql
 # TEMPORARY TOKEN (will be generated regularly and automatically in the future)
 NEXT_PUBLIC_LIBRARY_TOKEN=7f819e46e0eca2f5bebb642255a4cf555cf57220
 
+NEXT_PUBLIC_PUBHUB_BASE_URL=https://pubhub-openplatform.dbc.dk
```
ThomasGross, line 11: @spaceo should this be here?

components/pages/workPageLayout/WorkPageLayout.tsx
```diff
@@ -15,36 +15,42 @@ import {
...
+
+  const work = data?.work as WorkFullWorkPageFragment
+  const manifestations = work?.manifestations.all as ManifestationWorkPageFragment[]
+
+  // Filter out manifestations that not digital except physical books
+  const filteredManifestationsByMaterialType = filterManifestationsByMaterialType(manifestations)
+  // If multiple manifestations has the same type find the latest edition
+  const filteredManifestationsByEdition = filterManifestationsByEdition(
+    filteredManifestationsByMaterialType
+  )
```
spaceo, line 41: While I dont see any programmatic problems here, I think there is too much logic piling up.
Would be nice to separate functionality out in functions or hooks.
Another think we could consider was to use function composition, like described [here](https://medium.com/@karimdhrif4444/understanding-function-composition-in-javascript-2f0168d685ac) (and loads of other articles :))
And that way we could do, something like:
```js
const processManifestations = composer(filterManifestationsByMaterialType, filteredManifestationsByMaterialType)

const workManifestations = processManifestations(manifestations)

File: components/pages/workPageLayout/helper.ts [hunk 3]
```diff
@@ -1,30 +1,73 @@
...
+
+// Filter out manifestations that are not allowed material types
+export const filterManifestationsByMaterialType = (
+  manifestations: ManifestationWorkPageFragment[]
+) => {
+  return filter(manifestations, manifestation => {
+    // if (!manifestation.accessTypes?.length) {
+    //   return false
+    // }
+
```
spaceo, line 31: Should be removed I guess

components/pages/workPageLayout/helper.ts
```diff
@@ -1,30 +1,73 @@
...
+          item =>
+            item.materialTypes[0].materialTypeGeneral.code !==
+            current.materialTypes[0].materialTypeGeneral.code
+        )
+        acc.push(current)
+      }
+    }
+    return acc
+  }, [] as ManifestationWorkPageFragment[])
+}
```
spaceo, line 69: Could this be split into more functions? Seems overly complex. If not nevermind.

components/shared/workCard/WorkCard.tsx
```diff
@@ -84,6 +84,16 @@ const WorkCard = ({ work, className, isWithTilt = false }: WorkCardProps) => {
...
   )
+
+  // filter out manifestations that not digital except physical books
+  const filteredManifestationsByAccessType = filteredManifestations.filter(manifestation => {
+    if (manifestation.materialTypes[0].materialTypeGeneral.code === "BOOKS") {
+      return true
+    }
+
+    return manifestation.accessTypes?.[0].code !== "PHYSICAL"
+  })
```
spaceo, line 95: Why not moving this to the helper?

spaceo (approved):
Please look at my comments and consider cleaning up the code a bit.
Otherwise LGTM

PR #213, opened 2025-04-14
Do not use AP proxy on useLibraryTokenAlways services

spaceo:
#### Link to issue

#### Description

This PR should be merged after [this one](https://github.com/danskernesdigitalebibliotek/dpl-go/pull/208) has been reviewed because it contains a lot of the same commits.

The main focus in this PR is to make sure that Adgangsplatform services that are not dependent on an authorized user (aka only uses library token) do NOT use the ap service proxy, because we never need to attach the user token server side.
Right now the cover service is a service that only needs the library token and therefor is not being proxied.

But this PR also introduces a function for resolving the auth header in the AP service fetchers, so it is done in a uniform way.
And this PR makes sure that it is always possible to override the fetch headers from outside the fetchers.

### DIFF
--- a/app/ap-service/[[...slug]]/route.ts
+++ b/app/ap-service/[[...slug]]/route.ts
@@ -1,8 +1,7 @@
 import { NextRequest, NextResponse } from "next/server"
 
-import { TServiceType, getApServiceSettings } from "@/lib/helpers/ap-service"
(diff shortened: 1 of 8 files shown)


ThomasGross (approved):
LGTM

PR #215, opened 2025-04-15
Ddfbra 560 filtrer uonskede materialetyper ud i sogeresultat kald

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-560

#### Description

- This PR introduces changes that filter out unwanted material types in the search query
- NOTE: This change adds complexity to the graphql query and slows down the calls to FBI


components/shared/workCard/WorkCard.tsx
```diff
@@ -84,6 +84,16 @@ const WorkCard = ({ work, className, isWithTilt = false }: WorkCardProps) => {
     },
     [] as typeof manifestationsWithPublizonData
   )
+
+  // filter out manifestations that not digital except physical books
```
Adamik10, line 88: Wasn't there a similar function that you use somewhere else already that filters out the non-alowed types? Please use the same helper function where possible.

components/shared/workCard/WorkCard.tsx
```diff
@@ -99,18 +109,15 @@ const WorkCard = ({ work, className, isWithTilt = false }: WorkCardProps) => {
...
     return materialType?.materialTypeGeneral?.code === "PODCASTS"
   })
-  const isSomeManifestationTypeCostFree = filteredManifestations.some(
+
+  const isSomeManifestationTypeCostFree = filteredManifestationsByAccessType.some(
     manifestation => manifestation.publizonData?.costFree
   )
   // sort manifestations by materialTypeSortPriority
-  const sortedManifestations = filteredManifestations.sort((a, b) => {
+  const sortedManifestations = filteredManifestationsByAccessType.sort((a, b) => {
```
Adamik10, line 117: I think this sorting should also be somewhere globally accessible from a helper. We will need it for example on the search results to find out where to redirect to, right?

lib/config/resolvers/search.ts
```diff
@@ -43,6 +43,11 @@ const search = {
       filter: "materialTypesGeneral",
       translation: "Type",
     },
+    // Keep this for testing purposes
```
Adamik10, line 46: Very nice with a comment explanation :)

package.json
```diff
@@ -52,6 +52,7 @@
     "clsx": "^2.1.1",
     "cookies-next": "^5.1.0",
     "date-fns": "^4.1.0",
+    "fetch-to-curl": "^0.6.0",
```
Adamik10, line 55: Should this be removed after testing is done?

Adamik10 (commented):
I gotz some comments and questions :) Good job! 👏

PR #217, opened 2025-04-16
Load dpl-cms content by using new helper function

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-561

#### Description

Since /go and /go/[content-type] prefixes were added, we need functionality for resolving those paths in go.


Adamik10 (approved):
👍

PR #216, opened 2025-04-16
Fix quotas

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-547

#### Description
This PR:
- adds a new local Pubhub endpoint - v1/library/profile
- reworks the quota logic on user page (previous implementation was incorrect)

#### Screenshot of the result
-

#### Additional comments or questions
- reworking quotas when user is loaning will come in a separate PR

PR #219, opened 2025-04-16
Fix quotas when loaning materials

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-547

#### Description
This PR builds on top of #216 and fixes looking up quotas when loaning materials.

#### Screenshot of the result
-

-


ThomasGross (approved):
🙌

PR #221, opened 2025-04-23
Add support ID to profile page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-421

#### Description
This PR adds support ID for unilogin + bib users on the profile page.

#### Screenshot of the result
image

-

### DIFF
--- a/app/user/profile/ProfilePageLayout.tsx
+++ b/app/user/profile/ProfilePageLayout.tsx
@@ -8,6 +8,7 @@ import { ButtonSkeleton } from "@/components/shared/button/Button"
 import { userIsAnonymous } from "@/lib/helpers/user"
 import { getSession } from "@/lib/session/session"
 
+import SupportId, { SupportIdSkeleton } from "./SupportId"
 import Username, { UsernameSkeleton } from "./Username"
 
 const ProfilePageLayout = async () => {
@@ -20,7 +21,7 @@ const ProfilePageLayout = async () => {
   return (
     <div className="content-container grid-go -mt-space-y w-full space-y-3">
       <div className="col-span-full flex flex-row flex-wrap">
-        <h1 className="text-typo-subtitle-sm mb-5 lg:w-full">Profile</h1>
+        <h1 className="text-typo-subtitle-sm mb-5 opacity-50 lg:w-full">Profile</h1>
         <Suspense fallback={<ButtonSkeleton size="sm" />}>
           <LogoutButton />
         </Suspense>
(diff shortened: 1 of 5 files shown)


app/user/profile/ProfilePageLayout.tsx
```diff
@@ -20,14 +21,17 @@ const ProfilePageLayout = async () => {
   return (
     <div className="content-container grid-go -mt-space-y w-full space-y-3">
       <div className="col-span-full flex flex-row flex-wrap">
-        <h1 className="text-typo-subtitle-sm mb-5 lg:w-full">Profile</h1>
+        <h1 className="text-typo-subtitle-sm mb-5 opacity-50 lg:w-full">Profile</h1>
```
ThomasGross, line 24: This will also work and apply only on the text. Makes no different in this case though. :)

Remember to check darkmode when adding opacity. 50% looks very different with each background.

```suggestion
        <h1 className="text-typo-subtitle-sm mb-5 text-foreground/50 lg:w-full">Profile</h1>
```

app/user/profile/SupportId.tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (isLoading) {
+    return <SupportIdSkeleton />
+  }
+
+  if (!data?.userData?.friendlyCardNumber) {
+    return null
+  }
+
+  return (
+    <div className={cn("text-typo-subtitle-sm col-span-full -mt-5 pb-5 opacity-50", className)}>
```
ThomasGross, line 22: Same thing here: 

```suggestion
    <p className={cn("text-typo-subtitle-sm col-span-full -mt-5 pb-5 text-foreground/50", className)}>
```

app/user/profile/SupportId.tsx
```diff
@@ -0,0 +1,37 @@
...
+      {`Support ID: ${data.userData.friendlyCardNumber}`}
+    </div>
+  )
+}
+
+export const SupportIdSkeleton = ({ className }: SupportIdProps) => (
+  <div
+    className={cn(
+      "bg-background-skeleton -mt-5 mb-5 h-5 w-52 animate-pulse rounded-full pb-5",
```
ThomasGross, line 31: Consider if `-mt-5` is necessary here? Could we handle this differently?

app/user/profile/SupportId.tsx
ThomasGross, line 22: Consider if `-mt-5` is necessary here? Could we handle this differently?

ThomasGross (approved):
Looks good. Some small considerations added :)

PR #222, opened 2025-04-23
fix: material type filtering logic and remove unwanted material types

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-535

#### Description

Minor fixes to filter out unwanted materialtypes


Adamik10 (approved):
👏

PR #224, opened 2025-04-24
Solve weird cascade loading on front and search page

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-563

#### Description

The loading flow on the front page and on the search page was slow and seemed like it was waiting for requests to be fulfilled.
I found out that the functionality to get the librarytoken had turned into a server action and was called over and over again.

The architecture around the library token has now been restructured. It is still saved as a cookie so we can take advantage of the general principle of cookies expiring in browsers. But the value from the cookie is retrieved from the `getSession` function that is now a singe entrypoint for getting both the user and the library token. It means we do not have to struggle wit cookie fetching in a mixed client/server environment and we handle the user library token lookup in the same manner.

Also I excluded `ap-service` from the middleware to speed up the requests. That means the token maintenance functionality is being done on other routes - like the initial page load.

### DIFF
--- a/app/ap-service/[[...slug]]/route.ts
+++ b/app/ap-service/[[...slug]]/route.ts
@@ -18,14 +18,20 @@ const getAuthHeader = async (request: NextRequest, serviceType: TServiceType) =>
   const userToken = session?.adgangsplatformenUserToken
   const libraryToken = session?.adgangsplatformenLibraryToken
 
+  // If the settings (apServiceSettings) indicate that we should always use the library token,
+  // we will use the library token if it exists.
+  // Eg. the cover service always uses the library token because it does not need the user context.
   if (useLibraryToken && libraryToken) {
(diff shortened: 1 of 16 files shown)


app/ap-service/[[...slug]]/route.ts
```diff
@@ -11,11 +11,25 @@ const getAuthHeader = async (request: NextRequest, serviceType: TServiceType) =>
...
+  }
+
+  if (userToken) {
+    return `Bearer ${userToken}`
   }
+
+  if (libraryToken) {
+    return `Bearer ${libraryToken}`
+  }
+
```
Comment (ThomasGross) [inline app/ap-service/[[...slug]]/route.ts:32]: Im struggling to understand the logic of these if statements. Could we add a comment or two? 😊

ThomasGross (commented):
LGTM 👍

ThomasGross (approved):

PR #230, opened 2025-04-30
feat: refactor work card component to handle best representation manifestation + simplified workCard component

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-575

#### Description

This PR makes sure that the best representation manifestation is a available manifestation type.


spaceo (approved):
Nice with the cleanup!
Another time it would be nice if it is possible to separate the changes into multiple commits: One with cleanup and one with the changed behaviour.
But LGTM 👍

PR #231, opened 2025-04-30
feat: move fetching of config into a ContextProvider in the layout and update loginsheet

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-572

#### Description

This PR introduces includes:
- Handling of configuration data into a ContextProvider in the layout which makes config available for the client.


spaceo (changes requested):
Conf is separated into public and private.
And we should only insert configuration into the context provider that is allowed to be exposed to the client.
...and [this](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2349) should be merged first.

PR #234, opened 2025-05-01
Change go text body paragraph to use processed instead of value

Dresse:
#### Description

Currently the goTextBody GraphQL fragment is using "value" instead of "processed" to query the GoTextBodyParagraph.

We want to use "processed" instead, as we want the processed text from Drupal instead of the unprocessed version.

The reason for this is, that we have some WYSIWYG filters in Drupal which alters the text. 

One example of this is the LinkitFilter which makes sure to change all internal links to use the path alias instead of the internal "node/<id>" link. Instead they will look something like "artikel/<title-of-article>".

Also the PR updates the .gitignore file to include .idea files which are generated by PHPSTORM


lib/helpers/dpl-cms-content.ts
ThomasGross, line 1: Hvis /go/ prefixet er fjernet så skal vi ikke bruge getContentQueryPath funktionen længere 😊

ThomasGross (approved):
Lille kommentar.

Det ligner at storybook stories har ændret sig. Måske der skal gøres noget her? 

Det skyldes `value` også skal opdateres til `processed` i vores mockdata på `TextBody.stories.tsx`

lib/helpers/dpl-cms-content.ts
Dresse: Hov, det er en fejl fra min side, er fordi jeg har testet lokalt med at fjerne /go prefixed.
/go prefixet er ikke fjernet endnu. I hvertfald ikke før den anden opgave bliver merged ind. 

Skal jeg måske bare lave det om igen for nu og så kan i fjerne det fra alle de 3 indholdstyper når den anden opgave er merged/klar til at blive merged?

ThomasGross (changes requested):
Se kommentarer

Dresse:
> Se kommentarer

I updated the storybook story now, and now it works.

I also ran into trouble with the Cypress test, but I think that was because I had comitted the `"type": "module"` in package.json.

Please review again :)

ThomasGross (approved):
LGTM

PR #239, opened 2025-05-06
Link to parent library

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-505

#### Description
This PR adds the dynamic link to parent library into the header + footer.

#### Screenshot of the result
-

-

### DIFF
--- a/components/global/header/LinkToParentLibrary.tsx
+++ b/components/global/header/LinkToParentLibrary.tsx
@@ -1,0 +1,27 @@
+"use client"
+
+import React, { useContext } from "react"
+
+import SmartLink from "@/components/shared/smartLink/SmartLink"
+import { getEnv } from "@/lib/config/env"
+import { DplCmsConfigContext } from "@/lib/providers/DplCmsConfigContextProvider"
+
+const LinkToParentLibrary = () => {
+  const dplCmsConfig = useContext(DplCmsConfigContext)
+  const parentLibraryUrl = getEnv("DPL_CMS_HOSTNAME")
+
+  return (
+    <p className="text-typo-caption">
+      En del af{" "}
+      <SmartLink
+        className="animate-text-underline"
+        linkType="external"
+        href={parentLibraryUrl}
+        target="_blank">
+        {dplCmsConfig?.libraryInfo.name}
+      </SmartLink>
+    </p>
+  )
+}
+
+export default LinkToParentLibrary
--- a/components/shared/smartLink/SmartLink.tsx
+++ b/components/shared/smartLink/SmartLink.tsx
@@ -25,8 +25,9 @@ function SmartLink({
 
   // External link
   if (linkType === "external") {
+    const validHref = href.startsWith("http") ? href : `https://${href}`
     return (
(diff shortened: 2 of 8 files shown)


components/shared/smartLink/SmartLink.tsx
```diff
@@ -25,8 +25,9 @@ function SmartLink({
 
   // External link
   if (linkType === "external") {
+    const validHref = href.startsWith("http") ? href : `https://${href}`
     return (
-      <a className={className} href={href} target={target}>
+      <a className={className} href={validHref} target={target}>
```
spaceo, line 30: Just curious: Is it not possible to use an external url in the `next/link` component?

lib/config/dpl-cms/dplCmsConfig.ts
```diff
@@ -77,6 +80,9 @@ export const getDplCmsPublicConfig = async () => {
       logoutUrls: {
         adgangsplatformen: null,
       },
+      libraryInfo: {
+        name: "",
```
spaceo, line 84: I think we should make the default libraryInfo.name a nullable string (like the loginUrls/logoutUrls).
By doing that we are sure that when we are resolving the value we know that we did our best to retrieve the value, we did not succeed and mark it as `null`.
If we are using a string as our fallback value we do not know if we actually succeeded in retrieving the value, but for some weird reson it was not set in the dpl-cms.
Does it make sense?

components/global/header/LinkToParentLibrary.tsx
```diff
@@ -0,0 +1,27 @@
...
+      En del af{" "}
+      <SmartLink
+        className="animate-text-underline"
+        linkType="external"
+        href={parentLibraryUrl}
+        target="_blank">
+        {dplCmsConfig?.libraryInfo.name}
+      </SmartLink>
+    </p>
+  )
```
spaceo, line 24: What happens if we do not get the library name? As far as I can see that case is not being handled.

spaceo (changes requested):
Looks good. I have some commentz...

components/shared/smartLink/SmartLink.tsx
Adamik10: Here are some bullet points I found on why not 🚫  to do it:
	1.	No benefit: Next.js’s Link component adds client-side navigation, prefetching, and transition enhancements — none of which apply to external URLs.
	2.	Slight performance overhead: You’re wrapping something in a component whose behavior you’re not using.
	3.	No validation: Unlike internal links (which Next.js validates and optimizes), external ones are treated passively.

lib/config/dpl-cms/dplCmsConfig.ts
Adamik10: yesssssir :)

components/global/header/LinkToParentLibrary.tsx
Adamik10: I was expecting it to always be present, but good point. I changed it a bit now :)

spaceo (approved):
LGTM 👍

PR #242, opened 2025-05-06
Prevent build errors when fetching dplcms config

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-572

#### Description
Prevent build errors when fetching dplcms config
Make sure that error is catched so it does not block the build process


ThomasGross (approved):
Loooks good 😊

PR #244, opened 2025-05-07
Mock dpl-cms data and extend front page tests

JacobArrow:
#### Link to issue

[DDFBRA-524](https://reload.atlassian.net/browse/DDFBRA-524)

#### Description

This PR finalized my work with server side mocking in Nextjs, and extends the usage of fishery to mock server side requests to dpl-cms.
This PR also extends testing of the front page, including mocked data and navigation tests. 
Furthermore I've also introduced the ability to test on both desktop and mobile screen sizes.


.env.test
```diff
@@ -1,12 +1,10 @@
...
-
 # DPL-CMS
-NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS=https://route-to-dpl-cms.com/graphql
+NEXT_PUBLIC_DPL_CMS_HOSTNAME=http://localhost:9000
+NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_DPL_CMS=http://localhost:9000/graphql
 NEXT_PUBLIC_GRAPHQL_BASIC_TOKEN_DPL_CMS=Z3JhcGhxbF9jb25zdW1lcjp0ZXN0
 
 # FBI API
-NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_FBI=https://temp.fbi-api.dbc.dk/ereolgo/graphql
+NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_FBI=http://localhost:9000/graphql
```
spaceo, line 7: II think it would be nice to write a commet saying that localhost:9000 is our mock server. Setting the value here basically means that whenever we run some test in CI this will be used. Which is fine - but also nice to know :)

lib/config/dpl-cms/dplCmsConfig.ts
```diff
@@ -30,7 +30,7 @@ export const getDplCmsPrivateUniloginConfig = async () => {
 
   let configAPI = {}
   // If we are running in test mode, we don't want to query the DPL CMS config
-  const config = getEnv("TEST_MODE") ? null : await queryDplCmsPrivateConfig()
+  const config = await queryDplCmsPrivateConfig()
```
spaceo, line 33: Me so happy

next.config.mjs
```diff
@@ -1,6 +1,13 @@
 import withPlaiceholder from "@plaiceholder/next"
 import { env } from "process"
 
+function getImageHostname() {
+  if (env.NODE_ENV === "test") {
+    return "**"
+  }
```
spaceo, line 7: Is this still relevant? Could be that it is, just wnated to ask.

spaceo (approved):
LGTM 👍 
I have som considerations for you.

next.config.mjs
JacobArrow: Added some better comments on why that is used :)

PR #245, opened 2025-05-07
feat: add pubhub error messages to loanModal when requests are failing

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-507

#### Description

This PR introduces handling of error codes send through the publizon adapters and shows them in the loanModal. 

Screenshot 2025-05-08 at 09 16 49


app/(routes)/pubhub/v1/user/loans/[identifier]/route.ts
```diff
@@ -11,20 +11,28 @@ async function createdigitalLoan(
...
     }
   })
   try {
     const responseData = await createLoanRequest(context.uniLoginUserInfo, identifier)
-    return NextResponse.json(createLoan.parse(responseData))
+    const data = createLoan.parse(responseData)
+
+    // Check if the response code is not "101" (OK)
+    // https://docs.pubhub.dk/LibraryApi/2.7/Pubhub_Library_Integration_Guide_v2.7_September_2023.pdf
+    if (data.code !== "101") {
```
Comment (spaceo) [inline app/(routes)/pubhub/v1/user/loans/[identifier]/route.ts:31]: Nice with the comment. I think it would make sense to use the response code from the [generated code](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/5b777a0f5a7b0ef554c54cc503f9ba35dfb54d01/lib/rest/publizon/local-adapter/generated/model/apiResponseCode.ts#L17) to make it clear what it is.

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -63,7 +66,7 @@ const LoanMaterialModal = ({
   const { mutate } = usePostV1UserLoansIdentifier()
   const isbns = getIsbnsFromManifestation(manifestation)
   const [isHandlingLoan, setIsHandlingLoan] = useState(false)
-  const [error, setError] = useState<Error | null>(null)
+  const [publizonError, setPublizonError] = useState<{ code: number; message: string } | null>(null)
```
spaceo, line 69: Again to make it clear what a code is, I think it would make sense again to use the generated [ApiResponseCode](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/5b777a0f5a7b0ef554c54cc503f9ba35dfb54d01/lib/rest/publizon/local-adapter/generated/model/apiResponseCode.ts#L16C14-L16C29).

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -96,6 +103,21 @@ const LoanMaterialModal = ({
...
+  }
+  if (isErrorLoans || isErrorLibraryProfile) {
+    errors.push(
+      "Vi kunne desværre ikke bekræfte, om du kan låne materialet. Prøv venligst igen senere."
+    )
+  }
+  if (publizonError) {
+    errors.push(publizonErrorMessageMap[publizonError.code] || publizonError.message)
+  }
+
```
spaceo, line 120: I think it could make sense to move the errors logic in separate helper function. The `LoanMaterialModal` is way too big in my taste. We cannot write unit tests for the various parts and I honestly start to get worried if I get a task where I need to work on this part of the application.

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -122,30 +144,28 @@ const LoanMaterialModal = ({
...
-              <span className="font-bold">
-                {`"${getManifestationMaterialTypeSpecific(manifestation)}"`}
-              </span>{" "}
-              i denne måned.
-            </>
+        <div className="mx-auto mt-10 mb-5 w-full max-w-prose space-y-4">
+          <p className="text-typo-subtitle-md text-center">
+            {!isLoadingLibraryProfile &&
+              !isLoadingLoans &&
+              isLoanPossible &&
```
spaceo, line 151: could we move `!isLoadingLibraryProfile`and `! isLoadingLoans` into the logic of `isLoanPossible` so we just end up with `isLoanPossible`?

spaceo: You are checking for `errors`later on in the code. Shouldn´t it be part of `isLoanPossible` as well?

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -167,7 +187,9 @@ const LoanMaterialModal = ({
           onClick={() => closeModal()}>
           {!isLoadingLibraryProfile &&
             !isLoadingLoans &&
-            (!isLoanPossible || error || isErrorLibraryProfile || isErrorLoans ? "Nej" : "Luk")}
+            (!isLoanPossible || !errors.length || isErrorLibraryProfile || isErrorLoans
```
spaceo, line 190: Again can we consolidate everything into `isLoanPossible`?

components/shared/loanMaterialModal/helper.ts
```diff
@@ -0,0 +1,27 @@
+export const publizonErrorMessageMap = {
```
spaceo, line 1: We should use `ApiResponseCode`keys here as well

lib/rest/publizon/local-adapter/mutator/fetcher.ts
```diff
@@ -34,6 +34,8 @@ export const fetcher = async <ResponseType>({
 
     if (!response.ok) {
       console.error(response.status, response.statusText, serviceUrl)
+      const data = await response.json()
+      throw Error(JSON.stringify(data))
```
spaceo, line 38: Would be nice to move all code that are shared between the two publizon fetchers into one place. So we are sure that we have the same behaviour.

app/(routes)/ap-service/[[...slug]]/route.ts
```diff
@@ -70,18 +70,6 @@ async function proxyRequest(
...
-    console.log({ authHeader })
-    // eslint-disable-next-line no-console
-    console.log({ proxiedHeaders })
-    // eslint-disable-next-line no-console
-    console.log({ serviceUrl })
-    // eslint-disable-next-line no-console
-    console.log({ method })
-    // eslint-disable-next-line no-console
-    console.log({ body })
-
```
Comment (spaceo) [inline app/(routes)/ap-service/[[...slug]]/route.ts:84]: Nice. Away with that!

spaceo (changes requested):
Nice work. 🎉  I have some remarks.

spaceo (approved):
LGTM 👍

PR #251, opened 2025-05-13
fix: opacity issue on loanCards

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-608


Adamik10 (approved):
LGTM ✨

PR #258, opened 2025-05-19
Ddfbra 613 udvid lokal adapter til at hente bruger data gennem nye saop kald pa unilogin brugere

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-613

#### Description

This PR introduces the following enhancements:
- **Consistent Data Return**: Ensures that the ⁠getLibraryUserOrderListRequest in the local adapter always returns data. This change aims to prevent any potential breaks in the frontend functionality. 
- **Fallback Default Object**: In cases where the data returned by ⁠getLibraryUserOrderList is empty, a fallback default object has been implemented. This object includes a specific call to ⁠getFriendlyCardNumber, ensuring that necessary information is still accessible.

### DIFF
--- a/app/(routes)/pubhub/v1/user/loans/(lib)/requests.ts
+++ b/app/(routes)/pubhub/v1/user/loans/(lib)/requests.ts
@@ -50,19 +50,20 @@ export const getLibraryUserOrderListRequest = async (uniLoginUserInfo: TUserInfo
     cardnumber: uniLoginUserInfo.uniid,
   })
 
-  const friendlyCardNumberClient = await friendlyCardNumberClientAsync(
-    "./lib/soap/publizon/v2_7/wsdl/getfriendlycardnumber.wsdl"
-  )
-  const [friendlyCardNumberResponse] = await friendlyCardNumberClient.GetFriendlyCardnumberAsync({
-    ...getPublizonServiceParameters(),
-    cardnumber: uniLoginUserInfo.uniid,
-  })
-
   let libraryUserOrderList =
     libraryUserResponse.GetLibraryUserOrderListResult as TLibraryUserOrderList
 
   // Check if the response data is empty
+  // TODO: write a test for this case inside __tests__/pubhub.test.ts
   if (!libraryUserOrderList.response?.data) {
+    const friendlyCardNumberClient = await friendlyCardNumberClientAsync(
+      "./lib/soap/publizon/v2_7/wsdl/getfriendlycardnumber.wsdl"
+    )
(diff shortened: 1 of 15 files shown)


app/(routes)/pubhub/v1/user/loans/(lib)/requests.ts
```diff
@@ -1,16 +1,81 @@
...
   )
-  const [response] = await client.GetLibraryUserOrderListAsync({
+  const [libraryUserResponse] = await client.GetLibraryUserOrderListAsync({
     ...getPublizonServiceParameters(),
     cardnumber: uniLoginUserInfo.uniid,
   })
-  return response.GetLibraryUserOrderListResult as TLibraryUserOrderList
+
+  let libraryUserOrderList =
+    libraryUserResponse.GetLibraryUserOrderListResult as TLibraryUserOrderList
```
spaceo, line 54:
```suggestion
  const libraryUserOrderList =
    libraryUserResponse.GetLibraryUserOrderListResult as TLibraryUserOrderList
```
becasue of later suggestion (not tested)...

app/(routes)/pubhub/v1/user/loans/(lib)/requests.ts
```diff
@@ -1,16 +1,81 @@
...
+    const friendlyCardNumberResult =
+      friendlyCardNumberResponse.GetFriendlyCardnumberResult as TFriendlyCardNumberResultSchema
+
+    if (friendlyCardNumberResult.response.data) {
+      const friendlyCardNumber = friendlyCardNumberResult.response.data.FriendlyCardNumber
+
+      const tempObject = { ...defaultData }
+      tempObject.response.data.friendlycardnumber = friendlyCardNumber
+      libraryUserOrderList = defaultData
+    }
```
spaceo, line 77:
```suggestion
    if (friendlyCardNumberResult.response.data) {
      const friendlycardnumber = friendlyCardNumberResult.response.data.FriendlyCardNumber
      return { ...defaultData, friendlycardnumber   }
    }
```

spaceo: Note: not tested

spaceo (approved):
LGTM 👍 , but please look at comments

app/(routes)/pubhub/v1/user/loans/(lib)/requests.ts
ThomasGross: This would not overwrite the friendlycardnumber in the correct place. It would instead add friendlycardnumber in the root of the object.

PR #259, opened 2025-05-19
Quotas and blue titles - fix

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-587
https://reload.atlassian.net/browse/DDFBRA-589
https://reload.atlassian.net/browse/DDFBRA-584

#### Description
- Manually count loans in the quota section in user profile
- We used to rely on the data coming from Publizon, but we found out that this data isn't accurate with counting/not counting blue titles.
- Deletes local checking of quotas when loaning materials - we are relying on the service error response messages from now on

#### Screenshot of the result
-
-


app/(pages)/user/profile/LoanCard.tsx
```diff
@@ -50,6 +61,35 @@ const LoanCard = ({ manifestation, title, className }: LoanCardProps) => {
...
+            : [...prev, manifestationIsbn || "unknown isbn"]
+        )
+      }
+      if (isManifestationEbook(manifestation as ManifestationWorkPageFragment)) {
+        setEbookLoans(prev =>
+          prev.includes(String(manifestationIsbn))
+            ? prev
+            : [...prev, manifestationIsbn || "unknown isbn"]
+        )
+      }
```
ThomasGross, line 87: I find the logic in the ⁠LoanCard component to be quite confusing, especially with the addition of passing down ⁠setEbookLoans and ⁠setAudioLoans as methods. I believe that much of this logic should be managed in the parent component, which would simplify the ⁠LoanCard and make it easier to maintain and understand.

ThomasGross: But its working so maybe a comment with a TODO would be okay?

ThomasGross (changes requested):

ThomasGross (approved):

components/shared/loanMaterialModal/LoanMaterialModal.tsx
```diff
@@ -93,36 +81,12 @@ const LoanMaterialModal = ({
...
-    enabled: isbns.length > 0,
-  })
-
-  const isCostFree = publizonData?.product?.costFree || isManifestationPodcast(manifestation)
-  // Check if the user can loan more e-materials or if the material is cost-free (blue title or podcast)
-  const isLoanPossible =
-    (isCostFree && canUserLoanMoreCostFreeMaterials(dataLoans)) ||
-    canUserLoanMoreMaterials(dataLoans, dataLibraryProfile, manifestation)
-
   const errors: string[] = []
```
ThomasGross, line 84: We no longer need the error object. Consider using ⁠publizonError.code directly instead.

ThomasGross (approved):

PR #261, opened 2025-05-21
Dpl cms graphql basic token resolving in GO

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-628

#### Description

Resolve dpl cms graphql basic token in Go

It is easier to handle the basic auth token resolving in Go rather than
having the Dpl Platform doing it.


Dresse (approved):
LGTM 👍

PR #264, opened 2025-05-21
fix: changes to the error messages

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-567


Adamik10 (approved):
👏

PR #265, opened 2025-05-21
Move some configuration to dpl-cms

spaceo:
#### Link to issue

#### Description

Since we need non-generated secrets to come from the cms we move them to the go configuration query.
Also we need to make sure that values are overidable by the use of env variables for testing and development purposes.


ThomasGross (approved):
LGTM

PR #266, opened 2025-05-21
DDFBRA-144-test-sign-in-and-out

JacobArrow:
The openid client has some very sctrict rules for discovering af wellKnownUrl. But we can curcumvent these by setting the allowInsecureRequests property while in test mode

#### Link to issue

#### Description

Please include a short description of the suggested change and the reasoning behind the approach you have chosen.

#### Screenshot of the result

If your change affects the user interface you should include a screenshot of the result with the pull request.

If you have any further comments or questions for the reviewer them please add them here.

### DIFF
--- a/components/shared/sheet/LoginSheet.tsx
+++ b/components/shared/sheet/LoginSheet.tsx
@@ -1,5 +1,6 @@
 import React, { useContext } from "react"
 
+import { cyKeys } from "@/cypress/support/constants"
 import { DplCmsConfigContext } from "@/lib/providers/DplCmsConfigContextProvider"
 import { sheetStore } from "@/store/sheet.store"
 
@@ -25,7 +26,10 @@ function LoginSheet({ open }: { open: boolean }) {
                 Log ind med UNI•Login
               </div>
               <div>
-                <LoginButton url="/auth/login/unilogin" />
+                <LoginButton
+                  url="/auth/login/unilogin"
+                  data-cy={cyKeys["login-sheet-unilogin-button"]}
+                />
               </div>
             </div>
 
@@ -39,7 +43,12 @@ function LoginSheet({ open }: { open: boolean }) {
                   Login via Bibliotekernes fælles loginside
                 </div>
                 <div>
-                  {loginUrlAdgangsplatformen && <LoginButton url={loginUrlAdgangsplatformen} />}
+                  {loginUrlAdgangsplatformen && (
+                    <LoginButton
+                      url={loginUrlAdgangsplatformen}
+                      data-cy={cyKeys["login-sheet-adgangsplatformen-button"]}
+                    />
+                  )}
                 </div>
               </div>
             </>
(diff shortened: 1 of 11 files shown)


app/(pages)/user/profile/page.tsx
```diff
@@ -1,12 +1,15 @@
...
 
 import { setPageMetadata } from "@/lib/helpers/helper.metadata"
 
 import ProfilePageLayout from "./ProfilePageLayout"
 
 export const metadata: Metadata = setPageMetadata("Din profil")
 
-const Page = () => {
+const Page = async () => {
+  await connection()
```
ThomasGross, line 11: Just curious, why is the ⁠await connection() necessary in this context?

As I understand it, it seems to be required only when using client-specific functions.

.github/workflows/e2e-test.yml
```diff
@@ -34,12 +34,20 @@ jobs:
...
+        env:
+          # A bug in the current version of Nextjs causes NODE_ENV=test to brick the build
+          # By overridding the variables here we can ensure that the correct urls are set
+          TEST_MODE: true
+          NEXT_PUBLIC_DPL_CMS_HOSTNAME: http://localhost:9000
+          NEXT_PUBLIC_DPL_CMS_GRAPHQL_URL: http://localhost:9000/graphql
+          NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_FBS: http://localhost:9000/graphql
+          UNILOGIN_CLIENT_ID: http://localhost:9000/client
+          UNILOGIN_WELLKNOWN_URL: http://localhost:9000
+          NEXT_PUBLIC_APP_URL: http://localhost:3000
```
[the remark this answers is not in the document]
JacobArrow: This would not force Nextjs to actually use .env.cypress
Nextjs will always use `NODE_ENV` to decide which .env file is loaded
https://nextjs.org/docs/pages/guides/environment-variables#environment-variable-load-order

app/(pages)/user/profile/page.tsx
JacobArrow: I wrote this in my commit message:

> While building the application I noticed that Next is statically rendering the profile page, but as this page has logic to fetch cookies while checking for a valid session, it breaks the page while rendering. With `await connection()` we can force the page to be dynamic without using "force-dynamic".

components/shared/sheet/LoginSheet.tsx
```diff
@@ -27,7 +28,10 @@ function LoginSheet({ open }: { open: boolean }) {
                 Log ind med UNI•Login
               </div>
               <div>
-                <LoginButton url="/auth/login/unilogin" />
+                <LoginButton
+                  url="/auth/login/unilogin"
```
spaceo, line 32: Not that you have to implemented it right now, but this route should be defined like our other internal routes in [routes.ts](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/d837bc3f60ad33bb174ca3a497eb30c5a3abfb91/lib/config/resolvers/routes.ts#L1-L12).
Consider fixing it right a way or write a TODO otherwise

PR #272, opened 2025-05-22
Adjust the usage of the suspense wrapper to prevent layout shifts and ensure that certain components do not wait for irrelevant requests from ⁠dpl-cms before rendering

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-639

#### Description

This PR introduces an adjustment of the usage of the suspense wrapper to prevent layout shifts and ensure that certain components do not wait for irrelevant requests from ⁠dpl-cms before rendering.

### DIFF
--- a/lib/providers/DplCmsConfigContextProvider.tsx
+++ b/lib/providers/DplCmsConfigContextProvider.tsx
@@ -3,17 +3,32 @@
 import { createContext } from "react"
 
 import { TDplCmsPublicConfig } from "../config/dpl-cms/configSchemas"
+import { useGetDplCmsPublicConfigurationQuery } from "../graphql/generated/dpl-cms/graphql"
 
 export const DplCmsConfigContext = createContext<TDplCmsPublicConfig | null>(null)
 
-export default function DplCmsConfigContextProvider({
-  children,
-  config,
-}: {
-  children: React.ReactNode
-  config: TDplCmsPublicConfig | null
-}) {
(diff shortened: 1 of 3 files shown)


lib/providers/DplCmsConfigContextProvider.tsx
```diff
@@ -3,17 +3,32 @@
...
+          adgangsplatformen: data.goConfiguration.public.logoutUrls?.adgangsplatformen ?? null,
+        },
+        libraryInfo: {
+          name: data.goConfiguration.public.libraryInfo?.name ?? null,
+        },
+        unilogin: {
+          municipalityId: data.goConfiguration.public.unilogin?.municipalityId ?? null,
+        },
+      }
+    : null
```
spaceo, line 27: You are replicating what is happening [here](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/325bd14fa1c0a114af3c60abd50ea4828d7c6715/lib/config/dpl-cms/dplCmsConfig.ts#L98) you should use `getDplCmsPublicConfigData()` instead

spaceo (changes requested):
Very nice! 👏 
One thing needs to be changed though.

lib/providers/DplCmsConfigContextProvider.tsx
spaceo: I think also you need to handle suspense on the preview page. One Cypress test is failing...

spaceo (approved):
LGTM 👍

PR #280, opened 2025-05-27
Use lagoon env variables for cms routes

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-582

#### Description
Use lagoon env variables for cms routes
In the docker file we can make use of LAGOON_ROUTE which always point to the cms. And LAGOON_DOMAIN seems to be the same without the protocol.


ThomasGross (approved):
LGTM ✅

PR #282, opened 2025-05-28
feat: add handling of dynamic search profiles for fbi requests + env cleanup

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-657

#### Description

- add handling of dynamic search profiles for fbi requests
- changed `NEXT_PUBLIC_GRAPHQL_SCHEMA_ENDPOINT_FBI` to `CODEGEN_GRAPHQL_SCHEMA_ENDPOINT_FBI` as its only used in codegen context

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -15,10 +15,6 @@ UNILOGIN_WELLKNOWN_URL=https://hi-i-am-well-known-url.com
 # GO SESSION
 GO_SESSION_SECRET=yZJhgvvvUYMawDRbdzfdvsVswWXCKwkU
 
-# Only needed for FBI API codegen
-CODEGEN_GRAPHQL_SCHEMA_ENDPOINT_FBI=https://fbi-api.dbc.dk/ereolgo/graphql
-CODEGEN_LIBRARY_TOKEN=58bd3ff7782dd7bd678ca586d7b5f9fdfae85b58
-
 NEXT_PUBLIC_APP_URL=http://localhost:3000
 
 # PUBHUB


lib/helpers/ap-service.ts
```diff
@@ -8,12 +9,25 @@ export const getApServiceSettings = (serviceType: TServiceType) => {
...
 
-export const getApServiceUrl = (serviceType: TServiceType) => {
+export const getApServiceUrl = async (serviceType: TServiceType) => {
   const serviceSettings = getApServiceSettings(serviceType)
   if (!serviceSettings) {
     throw new Error(`No url found for the ${serviceType} service`)
   }
 
+  if (serviceType === "fbi") {
+    const dplCmsConfig = await getDplCmsPublicConfig()
```
spaceo, line 19: Since we have defaults (null) for the other settings, if we do not get them from the cms we should also add searchProfiles to [that](https://github.com/danskernesdigitalebibliotek/dpl-go/blob/6c984e6ebbc6d8c6f0278400175196af5e38515e/lib/config/dpl-cms/dplCmsConfig.ts#L83-L96)

lib/helpers/ap-service.ts
```diff
@@ -8,12 +9,25 @@ export const getApServiceSettings = (serviceType: TServiceType) => {
...
+
+    if (!dplCmsConfig.searchProfiles.local) {
+      throw new Error("DPL CMS searchProfiles.local is not defined")
+    }
```
spaceo, line 23: Are we sure we want to fail hard here or just log it? I am just asking because I am not sure :)

spaceo (changes requested):
Nice! 
I just have one thing I think needs to be added.
And one question :)...

PR #288, opened 2025-06-06
Use cover object to generate `srcset` to load directly from FBI

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-700

#### Description

This PR changes the image element in `CoverPicture.tsx` to use a simple `` element instead of the cached <Image>.
By doing this, we also need to generate our own `srcset`, by using the different size variants coming from FBI. 
This also requires the component to receive the full covers object.

### DIFF
--- a/components/shared/coverPicture/CoverPicture.tsx
+++ b/components/shared/coverPicture/CoverPicture.tsx
@@ -1,30 +1,21 @@
 import { useWindowSize } from "@uidotdev/usehooks"
 import { motion } from "framer-motion"
-import Image from "next/image"
 import React, { useEffect, useRef, useState } from "react"
 import Tilt from "react-parallax-tilt"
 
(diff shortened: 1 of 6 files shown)


ThomasGross (approved):
LGTM 🙌

components/shared/coverPicture/CoverPicture.tsx
```diff
@@ -50,41 +41,31 @@ export const CoverPicture = ({
...
-
-                setImageHeight(naturalHeight)
-                setImageWidth(naturalWidth)
-              }}
             />
           )}
-          {src && (
-            <Image
-              src={src}
+          {covers.large && (
```
ThomasGross, line 62: Is this check still relevant as we use srcSet?

ThomasGross (approved):

PR #289, opened 2025-06-07
Fix cookie saving in middleware

spaceo:
#### Description

Cookie was not saved because the request and response objects were not passed by reference properly to the `getIronSession` function.

The problem must have been introduced in the work with the dynamicIO/cache refactoring.


Adamik10 (approved):
👏 👏 👏

PR #291, opened 2025-06-08
Move caching to setup-node step

spaceo:
Move caching to setup-node step
The other method did not seem to work


JacobArrow (approved):
I'm approving this to get going, but I really think you should add some more context to your commit, as this change does not say anything on its own.

PR #294, opened 2025-06-10
Ddfbra 705 los hydration fejl i next

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-705

#### Description

[fix: moved html and body outside the suspense boundary](https://github.com/danskernesdigitalebibliotek/dpl-go/commit/954a97f62ca23ec0c2f3290df289d543a855d645)

[refactor: extract CategorySlide component for to fix hydration errors…](https://github.com/danskernesdigitalebibliotek/dpl-go/commit/b22fed80fa2bb6d7edba281a811b28114b017789) 

[fix: remove unnecessary height and width classes from layout div](https://github.com/danskernesdigitalebibliotek/dpl-go/commit/9d96047a282cc31df8d013bf8635aed91caf6fee)

[fix: height "auto" console error](https://github.com/danskernesdigitalebibliotek/dpl-go/commit/861feba24d219ccff6d21bb2f4a55c1624e47131)


components/shared/categorySlider/CategorySlider.tsx
```diff
@@ -69,19 +69,18 @@ function CategorySlider({ categories }: { categories?: TNodeGoCategory[] }) {
...
+                "has-checked:rotate-5",
+                "has-checked:-rotate-2",
+                "has-checked:rotate-3",
                 "has-checked:-rotate-6",
+                "has-checked:-rotate-5",
               ]
-
-              const randomIndex = Math.floor(Math.random() * rotations.length)
+              const randomIndex = index % rotations.length
               const isSelected = pathname === category.path
```
JacobArrow: WTF is going on here 😶
Seems like a very magic solution.

components/shared/categorySlider/CategorySlider.tsx
JacobArrow, line 1: I would appreciate a more explanatory commit message on why this change was needed.
I know it has something to do with Nextjs not liking Math.Random while rendering, and that useEffect will only run on the client, but this information should be available in the commit.

ThomasGross: Noted. Good point.

PR #299, opened 2025-06-16
Use potential www in go urls in startup script

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-719


ThomasGross (approved):
Looks good

PR #308, opened 2025-06-17
Get rid of prerender error on category slider

spaceo:
#### Description

The only way I can get rid of the prerendering error is by moving the category fetching to the category slider component itself.

The error before the fix:
image


codegen.ts
```diff
@@ -49,7 +49,7 @@ const config: CodegenConfig = {
     },
     // "lib/graphql/generated/dpl-cms/graphql.schema.json": {
     //   // TODO: Make this configurable
-    //   schema: "http://dapple-cms.docker/graphql",
+    //   schema: "http://dapple-cms.local/graphql",
```
ThomasGross, line 52: Please revert this until we decide to go with `.local`

```suggestion
    //   schema: "http://dapple-cms.docker/graphql",
```

ThomasGross (approved):
Please fix the minor comment. But all other looks good.

Adamik10 (approved):
Let's get this sucker in!

PR #312, opened 2025-06-24
feat: add guide to test production DB locally

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-693


ThomasGross:
> We should really have a task cmd for updating consumer pwd. And maybe a task cmd that then runs all three confugure cmds.
> 
> But anyways nice work :) LGTM 👍

Im gonna create a ticket addressing that.

PR #314, opened 2025-06-25
feat: change infobox styling to highlight the relevant issue more

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-734

#### Description

- changed infobox styling to highlight the relevant issue more


spaceo:
@ThomasGross Could you please provide a before/after screenshot and maybe also a url we it can be seen?

PR #333, opened 2025-06-26
Ddfbra 729 opsaetning af dependabot

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-729

#### Description

- Added dependabot groupings 
- Changed the next.js version to a fixed version number

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,53 @@ updates:
     directory: "/"
     # The maximum number of pull requests for version updates
     open-pull-requests-limit: 10
+    groups:
+      radix-ui:
+        patterns:
+          - "@radix-ui/*"
+      graphql:
+        patterns:
+          - "graphql"
+          - "@graphql-codegen/*"
+          - "@0no-co/graphqlsp"
+      eslint:
+        patterns:
+          - "eslint"
+          - "eslint-*"
+          - "@eslint/*"
+      typescript:
+        patterns:
+          - "typescript"
+          - "tsx"
+          - "@types/*"
+      prettier:
+        patterns:
+          - "prettier"
+          - "@prettier/*"
+      tailwindcss:
+        patterns:
+          - "tailwindcss"
+          - "@tailwindcss/*"
+          - "prettier-plugin-tailwindcss"
+          - "tailwindcss-animate"
+          - "tailwind-merge"
+          - "class-variance-authority"
+          - "clsx"
+      storybook:
+        patterns:
+          - "storybook"
+          - "@storybook/*"
+          - "@chromatic-com/storybook"
+      xstate:
+        patterns:
+          - "xstate"
+          - "@xstate/*"
+      vite:
+        patterns:
+          - "vite"
+          - "@vitejs/*"
+          - "vitest"
+
     # Check the npm registry for updates every day (weekdays)
     schedule:
       interval: weekly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,10 +24,8 @@ updates:
       react:
         patterns:
           - "react"
-          - "@types/react"
-      react-dom:
-        patterns:
           - "react-dom"
+          - "@types/react"
           - "@types/react-dom"
       next:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,4 @@
+# Dependabot configuration files that automate our dependency updates by creating pull requests.
 # The PRs are labeled and grouped by package ecosystem and semver update type.
 # We intentionally use grouping and pull request limits to avoid creating too many PRs and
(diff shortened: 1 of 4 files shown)


ThomasGross:
@kasperg Har du også mulighed for at tage et kig på groupings? Det kan være du gode tilføjelser eller nogle holdinger til struktur.

.github/dependabot.yml
```diff
@@ -10,6 +10,53 @@ updates:
...
     open-pull-requests-limit: 10
+    groups:
+      radix-ui:
+        patterns:
+          - "@radix-ui/*"
+      graphql:
+        patterns:
+          - "graphql"
+          - "@graphql-codegen/*"
+          - "@0no-co/graphqlsp"
```
kasperg, line 21:
```suggestion
      graphql-codegen:
        patterns:
          - "@graphql-codegen/*"
```

I suggest that you reduce this to a single group for `graphql-codegen`. The other packages do not have a direct relationship or dependency with each other - apart from the fact that they all deal with GraphQL.

The `@graphql-codegen/*` packages affect generated code so it makes sense to group them so we only have to re-generate code once but this is not the case with the other packages.

In other words:
- If the `graphql` page is updated nothing is likely to break if `@0no-co/graphqlsp` is not also updated.
 - If `@graphql-codegen/client-preset` is updated and `@graphql-codegen/introspection` there might be a dependency and we will like want to update both at the same time.

.github/dependabot.yml
```diff
@@ -10,6 +10,53 @@ updates:
...
+      eslint:
+        patterns:
+          - "eslint"
+          - "eslint-*"
+          - "@eslint/*"
+      typescript:
+        patterns:
+          - "typescript"
+          - "tsx"
+          - "@types/*"
```
kasperg, line 31: Again I do not see a reason to group these due to missing relationsships.

In general I would expect a `@types/*` packages to be placed in a group with the library it provide types for e.g. `@types/jwt-simple` and `jwt-simple`, `@types/react` and `@types/react-dom` with `react` and `react-dom` etc.

.github/dependabot.yml
```diff
@@ -10,6 +10,53 @@ updates:
...
+      prettier:
+        patterns:
+          - "prettier"
+          - "@prettier/*"
```
kasperg, line 35: Please reconsider having this group at all.

As far as I can tell there are no `@prettier/*` packages in `package.json` at the moment.

.github/dependabot.yml
```diff
@@ -10,6 +10,53 @@ updates:
...
+      tailwindcss:
+        patterns:
+          - "tailwindcss"
+          - "@tailwindcss/*"
+          - "prettier-plugin-tailwindcss"
+          - "tailwindcss-animate"
+          - "tailwind-merge"
+          - "class-variance-authority"
+          - "clsx"
```
kasperg, line 44: I think this group is too wide with many different libraries by different authors with the only common denominator being Tailwind.

.github/dependabot.yml
```diff
@@ -10,6 +10,53 @@ updates:
...
+          - "@chromatic-com/storybook"
+      xstate:
+        patterns:
+          - "xstate"
+          - "@xstate/*"
+      vite:
+        patterns:
+          - "vite"
+          - "@vitejs/*"
+          - "vitest"
```
kasperg, line 58: Hmmm. Vite is not even included as a package in `package.json`. Why is it included here?

ThomasGross:
@kasperg thanks for your comments. I've reviewed them all and restructured the groupings to be more precise. Please have another look.

.github/dependabot.yml
```diff
@@ -10,6 +10,79 @@ updates:
...
+          - "typescript"
+          - "tsx"
+      react:
+        patterns:
+          - "react"
+          - "@types/react"
+      react-dom:
+        patterns:
+          - "react-dom"
+          - "@types/react-dom"
```
kasperg, line 31: Please put these in one `react` group.

As I see it you usually use the same version of `react` and `react-dom`.

kasperg (approved):
:+1: I think you should join the `react` dependencies. Otherwise I think this is good to go.

.github/dependabot.yml
```diff
@@ -10,6 +10,77 @@ updates:
...
+          - "react-dom"
+          - "@types/react"
+          - "@types/react-dom"
+      next:
+        patterns:
+          - "next"
+          - "@next/env"
+          - "next-themes"
+          - "eslint-config-next"
+          - "@next/eslint-plugin-next"
```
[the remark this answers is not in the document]
ThomasGross: Yes im a little unsure how to handle this in terms of Dependabot grouping because next is in canary. Right now the next version is locked to at specific version in package.json. 

Do you suggest we remove the grouping?

PR #350, opened 2025-07-18
feat: add cypress tests for showing loans on profile page

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-224

#### Description

Cypress tests added for showing loans UI on profile page for both unilogin and adgangsplatform


cypress/e2e/login-logout.cy.ts
```diff
@@ -115,6 +118,39 @@ describe("UNI•Login: Login / Logout API Tests", () => {
...
+      statusCode: 200,
+      body: getV1ProductsIdentifierAdapterFactory.transient({ id: "9788711917141" }).build(),
+      headers: { "content-type": "application/json" },
+    })
+
+    cy.intercept("GET", "/ap-service/pubhub-adapter/v1/products/8788711917141", {
+      statusCode: 200,
+      body: getV1ProductsIdentifierAdapterFactory.transient({ id: "8788711917141" }).build(),
+      headers: { "content-type": "application/json" },
+    })
```
spaceo, line 146: Are these interceptions necessary? 
Since it is a Unilogin user there should not be any requests towards the pubhub-adapter, right?

ThomasGross: As its currently implemented (in the actual code and not only the test) unilogin and adgangsplatform users both uses the pubhub-adapter. In other words for this specific call the proxy is not used. 

If we want to show if the manifestation is a blue title, then this intercept method its required.

cypress/e2e/login-logout.cy.ts
```diff
@@ -136,11 +172,6 @@ describe.only("Adgangsplatformen: Login / Logout API Tests", () => {
       data: getAdgangsplatformenUserToken.build(),
     })
 
-    cy.interceptGraphql({
-      operationName: "complexSearchForWorkTeaser",
-      data: complexSearchForWorkTeaser.build(),
-    })
-
```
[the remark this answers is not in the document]
ThomasGross: I think this was a mistake. Still i did not see any issues on my tests. But it should be present, as we dont want to call the actual service.

cypress/e2e/login-logout.cy.ts
spaceo: Ah ja makes sense

spaceo (changes requested):
Looks good!
But a Cypress failing test needs to be fixed/rerun before it can be approved

cypress/e2e/login-logout.cy.ts
```diff
@@ -172,6 +160,14 @@ describe("UNI•Login: Login / Logout API Tests", () => {
 })
 
 describe("Adgangsplatformen: Login / Logout API Tests", () => {
+  beforeEach(() => {
+    mockFrontpage()
+    cy.visit("/")
+
+    // Ignore fetch errors from DPL CMS
+    cy.expectError("Failed to fetch data from DPL CMS")
```
spaceo, line 168: Same as above

spaceo (approved):
LGTM 👍

PR #371, opened 2025-08-15
fix: update npm packages to ensure none contain critical vulnerabilities + bump node version to v20

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-306

#### Description

- update npm packages to ensure none contain critical vulnerabilities 
- bump node version to v20


spaceo (approved):
In Reload we thrive to, if possible, to only make the changes needed in every commit.
Sometimes it is hard, if not possible.
But in this case, it would be easy to make an interactive rebase and undo the node update in the first commit.
I like the way Arne describes the idea:
In every commit try to only add what is needed and whar makes sense in the context.
Does it make sense?

Otherwise
LGTM 👍  🎉

PR #378, opened 2025-08-19
feat: implement material type redirection when no type is present in …

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-310

#### Description

The new Go solution must be able to handle redirects from the old ereolengo.dk site to the new one via https://reload.github.io/ereolen-go/

Material pages in the old solution do not have a type specified in the URL:
https://ereolengo.dk/ting/object/870970-basis%3A141030485?u_navigatedby=Nye%20eb%C3%B8ger

The new Go solution expects a type parameter in order to display a page:
https://go.bibliotek.kk.dk/work/work-of%3A870970-basis%3A27866360?type=AUDIO_BOOKS

To resolve this issue, the following logic has been implemented:
1. If a user navigates to a material page in Go without a type, we will set one via code.
2. The type to be set is the bestRepresentation manifestation type.
3. If bestRepresentation is not an allowed manifestation vi will fallback to the first allowed manifestation in manifestations list

### DIFF
--- a/components/pages/workPageLayout/WorkPageLayout.tsx
+++ b/components/pages/workPageLayout/WorkPageLayout.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { notFound, useSearchParams } from "next/navigation"
+import { notFound, useRouter, useSearchParams } from "next/navigation"
 import React, { useEffect, useMemo, useState } from "react"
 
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #386, opened 2025-08-22
Use wsdl-client fork

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-324

#### Description

Temporary solution, in order to get rid of log noise and ts errors. @arnested is working on a new fork in the Reload NPM organisation - when it has been released we can switch to that.


arnested:
It's ready: https://www.npmjs.com/package/@reloaddk/wsdl-tsclient

PR #403, opened 2025-08-27
Adding documentation of the cache handling in Go

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-210

#### Description

Documentation of the thoughts behind the cache handling in Go


docs/architecture/adr-008-caching-handling.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+GET /api/cache/revalidate?secret={DRUPAL_REVALIDATE_SECRET}&tags=tag1,tag2,tag3
+```
+
+#### Manual Invalidation
+
+```typescript
+// In React Query
+queryClient.invalidateQueries({ queryKey: ["search", searchTerm] })
```
kasperg, line 163: Consider if a clearer relationship between the example here and in "3rd Party Service Queries" can be make.

Should this use something like ` useSearchWithPaginationQuery.getKey(searchTerm)`?

docs/architecture/adr-008-caching-handling.md
```diff
@@ -0,0 +1,202 @@
...
+
+// Next.js cache tags
+revalidateTag("category-123")
+revalidatePath("/kategori/some-category")
```
kasperg, line 167: Do we use this at all?

kasperg (changes requested):
The information here is not wrong but as I read this and consider the discussions we have had on this point I do not think they are addressed in a clear manner.

Reading it I see two architecturally important decisions:

1. How we handle a situation with a decoupled CMS and frontend where we a) want editorial changes to take effect immediately and b) consider it performance/resource-wise unfeasible to call the CMS AP live for each page request (due to the choice of using GraphQL, resulting in poor CMS cachability and this high response times - but that is probably outside the scope of this ADR). This has lead us to deciding on an experimental version of Next, which in turn affects maintainability. Does the `next-drupal` package also warrant a mention here?
2. Caching requests to business systems for the sake of end user performance - even though they in the digital library infrastructure are intended to be called directly to avoid cache complexity in the frontend.

Even though they both deal with caching I consider 1. is a lot more important than 2. 

Perhaps it would be worth splitting this work into two different ADRs to make these different decisions stand out more clearly.

PR #404, opened 2025-08-27
Make sure that headers are sent to FBI in proper format

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-346

#### Description

Make sure that headers are sent to FBI in proper format
If we do not turn header list into an object in ends up being an array of arrays.


ThomasGross (changes requested):
LGTM! However, the end-to-end tests are failing. Please check if the Cypress tests are flaky due to this change. 👍

ThomasGross (approved):
LGTM 👍

PR #405, opened 2025-08-28
Use query key function for user loans

spaceo:
This is usually our way of defining React Query keys.

See the [dpl-react
repository](https://github.com/danskernesdigitalebibliotek/dpl-react) for more examples.


ThomasGross (approved):
LGTM 👍

PR #411, opened 2025-09-02
Do not use node middleware when checking health

spaceo:
In order to prevent unwanted side effects or log warning/errors we should disable the middleware in the heath route.


ThomasGross (approved):
LGTM 👍

PR #414, opened 2025-09-02
Dependabot ignore

ThomasGross:
#### Link to issue

none

#### Description

This PR will add ignores to the dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -89,6 +89,9 @@ updates:
         patterns:
           - "cypress"
           - "eslint-plugin-cypress"
+    ignore:
+      - dependency-name: "@types/node"
+        update-types: ["version-update:semver-major"]
 
   # Enable version updates for GitHub Actions
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


spaceo (commented):
I think there should be an explanation, preferably in the commit, otherwise in the PR, about why this is needed.

.github/dependabot.yml
```diff
@@ -89,7 +75,29 @@ updates:
...
           - "cypress"
           - "eslint-plugin-cypress"
-
+    ignore:
+      - dependency-name: "@types/node"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "next"
+        versions: [">=15.4.0-canary.47"]
+      - dependency-name: "@next/*"
+        versions: [">=15.5.2"]
```
kasperg, line 84: What is the relationship between `next` and `@next/*` versions? Should they be the same like it is the case for e.g. Storybook?

To me it seems to be the case since the both stem from the same monorepo at https://github.com/vercel/next.js.

If this is the case then I suggest pinning them to the same version e.g. `15.4.0` and ignoring further updates. Other `@next/*` packages are currently at `15.0.3`.

kasperg (changes requested):
:+1: Nice breaking these updates apart and pinning to current major version.

I think the next version definition needs a bit more thought before this can be merged.

PR #719, opened 2026-01-15
fix: cover sizing in GO material slider

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDF-30

#### Description

•  Fixed the issue where work covers in the GO material slider sometimes rendered very small on initial load and only resized correctly after a window resize.  
•  Switched CoverPicture from using window width–based measurements to a ResizeObserver that tracks the actual slide/container size, so cover sizing updates whenever Keen slider changes slide width.  
•  Implemented “fit inside container” logic for covers (equivalent to object-fit: contain): covers keep their aspect ratio, never overflow the card, and always use the maximum possible size within the container.

### DIFF
--- a/components/shared/coverPicture/CoverPicture.tsx
+++ b/components/shared/coverPicture/CoverPicture.tsx
@@ -129,7 +129,7 @@ export const CoverPicture = ({ covers, alt, withTilt = false, className }: Cover
           <Icon
             name="question-mark"
             className="text-foreground h-[50px] opacity-20 lg:h-[100px]"
-            aria-label="Question mark icon"
+            aria-label="Spørgsmålstegn ikon"
           />
           <p className="text-typo-caption text-center opacity-55">Billede kunne ikke vises</p>
         </motion.div>
--- a/components/shared/coverPicture/CoverPicture.tsx
+++ b/components/shared/coverPicture/CoverPicture.tsx
@@ -1,3 +1,5 @@
+"use client"
+
 import { motion } from "framer-motion"
 import React, { useEffect, useRef, useState } from "react"
 import Tilt from "react-parallax-tilt"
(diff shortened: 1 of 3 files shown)


components/shared/workCard/WorkCardStackedWithCaption.tsx
```diff
@@ -92,7 +92,6 @@ const WorkCardStackedWithCaption = ({
               title={title}
               bestRepresentation={manifestation}
               manifestations={manifestations}
-              isWithTilt
```
kasperbirch1, line 95: Are you sure about this? It seems like this style is used in other places as well and is part of their identity.

components/shared/coverPicture/CoverPicture.tsx
```diff
@@ -90,7 +129,7 @@ export const CoverPicture = ({ covers, alt, withTilt = false, className }: Cover
           <Icon
             name="question-mark"
             className="text-foreground h-[50px] opacity-20 lg:h-[100px]"
-            aria-label="Spørgsmålstegn ikon"
+            aria-label="Question mark icon"
```
kasperbirch1: Why is this in English? As far as I can see, it doesn’t seem to be translatable.

components/shared/coverPicture/CoverPicture.tsx
```diff
@@ -1,4 +1,5 @@
-import { useWindowSize } from "@uidotdev/usehooks"
+"use client"
```
kasperbirch1, line 1: I’m confused about why this worked without it before. Can it be inherited from the parent?

kasperbirch1 (approved):
My comments are more questions than suggestions for code changes, so I’ll approve. 👍

PR #750, opened 2026-01-27
Use go profile with FBI

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDF-34

#### Description

Go has its own search profile now. We'll hardcode it as it's not supposed to vary across sites.

This is not done, there's some codegen that needs to be run, but according to the docs it's:

```shell
yarn codegen:graphql
```

Well...

```shell
~/d/d/dpl-go ▶ yarn codegen:graphql
fish: Unknown command: yarn
```

And no `docker-compose.yml` to get a containerized environment. So I'm passing the buck to the frontenders.


ThomasGross (approved):
LGTM

PR #767, opened 2026-02-06
Upgrade Node 20=>22.

rasben:
We're going with 22, as that's what we're using in the CMS.

In the future, after mono repo, we'll look into upgrading everything to 24 (the current stable)


rasben:
Relies on

github.com/danskernesdigitalebibliotek/dpl-platform/pull/new/node-22

xendk:
> In the future, after mono repo, we'll look into upgrading everything to 25 (the current stable)

No, security support ends in three months. 24.

PR #835, opened 2026-03-16
Ddf 35 go viser nu ogsa billedboger graphic novels og tegneserier

ThomasGross:
# Do not open new pull requests here!

New pull requests should be created in dpl-web:

https://github.com/danskernesdigitalebibliotek/dpl-web


kasperbirch1 (approved):
I understand we are about to switch to dpl-web, but a little more explanation in the PR itself would be nice. I have relied on the explanation in the ticket: https://reload.atlassian.net/browse/DDF-35

Other than that, good job!

app/(pages)/user/profile/LoanCard.tsx
```diff
@@ -47,9 +44,9 @@ const LoanCard = ({
   const today = new Date()
   const daysUntil = differenceInDays(targetDate, today)
 
-  const isCostFree =
-    dataProducts?.product?.costFree ||
-    isManifestationPodcast(manifestation as ManifestationWorkPageFragment)
+  const materialTypeCode = manifestation.materialTypes[0]?.materialTypeSpecific.code
```
kasperbirch1, line 47: Just a note: 

We have talked about how using [0] directly can lead to unexpected bugs. If you have Lodash in the project, we should use _.first() (or _.head()) instead, because it handles null or undefined arrays more gracefully.

